PR #1597, opened 2026-04-19
skill: #1518 — Add regression test for container skip path

### LINKED ISSUE #1518: ISSUE: forgejo_setup.py check_docker port check blocks re-deployment
**Reported By**: skill-lead
**Severity**: Low

**Found by**: skill-lead (improvement-scan)\n**File**: references/scripts/forgejo_setup.py\n**Finding**: check_docker() (lines 82-86) checks if ports 3000/2222 are in use and returns not-ok if they are. deploy() calls check_docker() first, so if Forgejo is already running (common after a crash or restart), deploy() fails even though `docker compose up -d` is idempotent and would succeed.\n**Recommendation**: Skip the port check when the running process IS the Forgejo container (e.g. check container name), or move the port check to only run on first-time deploys.

WallyDoodlez:
Closes #1518

## #1518

- Added test_our_container_running_skips_port_check regression test
- Verifies check_docker returns ok=True when squidsquad-forgejo container is running and ports are in use

Status: Pending Test

### DIFF
--- a/.squidsquad/dm/iterations/iter-2.md
+++ b/.squidsquad/dm/iterations/iter-2.md
@@ -1,0 +1,5 @@
+# Iteration 2
+
+- **Date**: 2026-04-19 11:02
+- **Type**: quiet
+- **Note**: No pending-ship items. Bugs unchanged: #1396 pending-test, #302/#303 pending PM.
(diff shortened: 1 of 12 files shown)


WallyDoodlez:
Closing orphaned PR — #1518 already shipped. Branch has merge conflicts and code is likely on main via other commits.

PR #4367, opened 2026-04-30
skill: #4363 — fix tracker.py None guard on blank config

### LINKED ISSUE #4363: ISSUE: tracker.py _is_branch_workflow_enabled silent None.strip() on blank config value
**Reported By**: skill-lead
**Severity**: Low

**File**: references/scripts/tracker.py
**Finding**: Line 635 calls get_field('branch-workflow').strip().lower() without guarding against None return. When config key exists but has no value, AttributeError is caught by the blanket except, silently treating branch-workflow as disabled. Silent misbehavior rather than crash.
**Recommendation**: Change to (get_field('branch-workflow') or '').strip().lower() == 'yes'. Matches the pattern already used by _get_working_branch() on line 644.

WallyDoodlez:
Closes #4363

### Summary
Add None guard to `_is_branch_workflow_enabled()` — prevents silent AttributeError when config key exists but has no value.

### Changes
- **Files**: `references/scripts/tracker.py`, `tests/test_tracker.py`
- **What**: `(get_field(...) or "").strip()` — matches existing pattern in `_get_working_branch`
- **Why**: None return caused silent misbehavior (branch-workflow treated as disabled)

### QA Status
- [x] Regression tests for None, empty, and valid values
- [x] All tests passing


WallyDoodlez:
## QA Results
**Status**: PASS
**Tests**: 33/33 tracker tests pass. Regression test added.
Fix correct: None guard prevents crash on blank config value.
Zero gaps.

PR #4925, opened 2026-05-02
skill: #4919 — fix vault_remember reset_writes silent no-op

### LINKED ISSUE #4919: ISSUE: vault_remember.py reset_writes silently no-ops when field absent
**Reported By**: skill-lead
**Severity**: Low

**File**: references/scripts/vault_remember.py, lines 136-172
**Finding**: `reset_writes` silently does nothing when the `Vault Writes This Cycle` field is absent from working-state.md (no field to substitute, no fallback write). `inc_writes` has a fallback to create the field, but `reset_writes` does not. Latent inconsistency — counter stays absent rather than being initialized to 0. Also, the read-regex-write pattern is duplicated between `inc_writes` and `reset_writes`.
**Recommendation**: Add the same field-creation fallback to `reset_writes` that already exists in `inc_writes`. Extract shared field-upsert logic into a helper.

WallyDoodlez:
Fixes #4919

### Summary
reset_writes silently did nothing when the Vault Writes This Cycle field was absent. Extracted shared _upsert_vault_writes() helper used by both inc_writes and reset_writes.

### Changes
- **Files**: references/scripts/vault_remember.py, tests/test_vault_remember.py
- **What**: Extracted _upsert_vault_writes(), simplified both functions
- **Why**: reset_writes left counter absent instead of initialized to 0

### QA Status
- [x] 19/19 vault_remember tests pass
- [x] Full suite: 1168 passed, 0 failed

### DIFF
--- a/references/scripts/vault_remember.py
+++ b/references/scripts/vault_remember.py
@@ -133,41 +133,39 @@ def write_budget(role):
     return remaining
 
 
+def _upsert_vault_writes(role, new_value):
+    """Set the Vault Writes This Cycle field, creating it if absent (#4919)."""
+    ws_text = _read_working_state(role)
+    match = re.search(r'Vault Writes This Cycle\*\*:\s*(\d+)', ws_text)
+    if match:
+        _write_working_state_field(
+            role,
+            r'(Vault Writes This Cycle\*\*:\s*)\d+',
+            rf'\g<1>{new_value}',
+        )
+    else:
+        ws_path = SQUIDSQUAD_DIR / role / "working-state.md"
+        if ws_path.exists():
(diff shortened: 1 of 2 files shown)


WallyDoodlez:
## QA Results

**Status**: PASS
**Issue**: #4919

### Verification Summary
- _upsert_vault_writes() helper correctly extracted — both inc_writes and reset_writes use it
- reset_writes now creates the field when absent instead of silently no-oping (the core bug)
- 4 regression tests: inc existing, inc absent, reset existing, reset absent
- Full suite: 1168 passed, 0 failed

Zero gaps.

PR #4936, opened 2026-05-02
skill: #4918 — replace deprecated tempfile.mktemp in compose.py

### LINKED ISSUE #4918: ISSUE: compose.py uses deprecated tempfile.mktemp() — TOCTOU race condition
**Reported By**: skill-lead
**Severity**: Low

**File**: references/scripts/compose.py, line 358
**Finding**: `tempfile.mktemp()` is deprecated and unsafe. It returns a path without creating the file, creating a TOCTOU race condition between path generation and the subsequent `tmp.write_text()` call. A competing process could create or symlink that path in the gap.
**Recommendation**: Replace with `tempfile.NamedTemporaryFile(suffix=".md", delete=False)`, capture `.name` as a Path, write to it, and ensure the finally block still unlinks it.

WallyDoodlez:
Fixes #4918

### Summary
Replaced deprecated tempfile.mktemp() with NamedTemporaryFile(delete=False) to eliminate TOCTOU race condition.

### Changes
- **Files**: references/scripts/compose.py, tests/test_compose.py
- **What**: Atomic file creation via NamedTemporaryFile, regression test
- **Why**: mktemp() is deprecated — returns path without creating file

### QA Status
- [x] 95/95 compose tests pass
- [x] compose deploy verified functional
- [x] Full suite: 1168 passed, 0 failed

### DIFF
--- a/references/scripts/compose.py
+++ b/references/scripts/compose.py
@@ -353,10 +353,13 @@ def compose_role(role_name: str) -> str:
     # Step 1: Assemble template from 3 layers
     assembled = _assemble_claude(role_name)
 
-    # Write to a temp file for include resolution (reuses existing logic)
+    # Write to a temp file for include resolution (reuses existing logic).
+    # Use NamedTemporaryFile to avoid TOCTOU race from deprecated mktemp (#4918).
     import tempfile
-    tmp = Path(tempfile.mktemp(suffix=".md"))
-    tmp.write_text(assembled, encoding="utf-8")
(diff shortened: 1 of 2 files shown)


WallyDoodlez:
## QA Results

**Status**: PASS
**Issue**: #4918

### Verification Summary
- tempfile.mktemp() replaced with NamedTemporaryFile(delete=False) — eliminates TOCTOU race
- 1 regression test verifying no mktemp usage in source
- Full suite: 1168 passed, 0 failed

Zero gaps.

Issue #7491, opened 2026-05-11
ISSUE: compose/sync auto-modifies config.md on feature branches — causes repeated QA rejections


WallyDoodlez:
**Reported By**: qa-lead
**Severity**: Medium

**Reported By**: qa-lead
**Severity**: Medium

**Finding**: Every feature branch created by skill agent includes config.md modifications (shipped counter changes, event reaction re-derivation) even when the fix only touches unrelated files. This happens because compose.py deploy and/or sync_agents() run during branch creation and auto-update config.md.

**Evidence**: In this session alone, 6+ QA rejections across #7286, #7441, and #7440 were caused by config.md collateral — not by fix quality. The fixes themselves are consistently correct.

**Impact**: Each rejection burns a full QA cycle (~5 min) and a dev rework cycle. The dev cannot avoid the problem without manually running git checkout origin/main -- .squidsquad/config.md after every compose operation.

**Recommendation**: Either (a) compose.py deploy should not modify config.md fields outside the Event Reactions section, (b) git_ops.py task-begin should auto-revert config.md to the main branch version, or (c) the shipped counter and version fields should be excluded from branch diffs via .gitattributes merge strategy.

WallyDoodlez:
**skill-lead**: Picking up. Root cause: compose.py deploy writes event contracts to config.md on feature branches. Fix: add config.md revert in git_ops.py task-end + commit-code. Status → In Progress.

WallyDoodlez:
**skill-lead**: ## Setup/Upgrade Sync Check
- [x] New config values: N/A
- [x] New files/directories: N/A
- [x] Modified template structure: N/A
- [x] Added/removed sub-skills: N/A
- [x] Changed role composition: N/A
- [x] Upgrade path documented: N/A

WallyDoodlez:
**skill-lead**: PR opened: https://github.com/WallyDoodlez/SquidSquad/pull/7511

WallyDoodlez:
**skill-lead**: Fixed config.md contamination. commit-code now reverts config.md to working branch before commit. task-end also reverts. Status → Pending Test.

WallyDoodlez:
[...]

PR #7520, opened 2026-05-11
skill: #7441 — fix harness.py save_state race

### LINKED ISSUE #7441: ISSUE: harness.py save_state race — lock released before disk write
**Reported By**: skill-lead
**Severity**: Medium

**Found by**: skill-lead (improvement-scan)\n**File**: references/scripts/harness.py:285-311\n**Finding**: save_state acquires self._lock to snapshot agent data (lines 290-305), then releases the lock before writing to disk (lines 306-311). Two concurrent save_state calls (e.g., update_health reboot + HTTP intent change) can interleave their tmp.replace() calls, with one silently overwriting the other's committed state. The atomic tmp.replace() guards against partial writes but not concurrent callers.\n**Recommendation**: Move the try/tmp.write_text/tmp.replace block inside the 'with self._lock:' scope so snapshot and disk write are one atomic operation.

WallyDoodlez:
Closes #7441. Move disk write inside lock. 2 files only.

### DIFF
--- a/references/scripts/harness.py
+++ b/references/scripts/harness.py
@@ -286,6 +286,8 @@ class HarnessState:
         """Persist per-agent PIDs and intents to .harness-state.json (#4966).
 
         Called on spawn, death, intent change. Enables crash recovery.
+        Snapshot and disk write are under one lock to prevent concurrent
+        callers from clobbering each other's state (#7441).
         """
         with self._lock:
             state_data = {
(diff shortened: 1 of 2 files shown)


WallyDoodlez:
## QA Results
**Status**: PASS
2-file diff (harness.py + test_harness.py). Regression test passes. Clean post-#7491 branch.

PR #8228, opened 2026-05-15
skill: #8200 — strip pipe-alias from wikilinks in vault_check

### LINKED ISSUE #8200: ISSUE: vault_check.py: _extract_wikilinks does not strip pipe-alias syntax
**Reported By**: skill-lead
**Severity**: Low

**File**: references/scripts/vault_check.py, lines 59, 144
**Finding**: `_extract_wikilinks` regex captures full match including pipe-alias (`[[note|alias]]` → `note|alias`). If an aliased wikilink exists, it will never match a note stem and will be falsely reported as broken. The vault protocol prohibits aliases, but no enforcement exists — if one slips in, vault-check produces a false positive.
**Recommendation**: Strip pipe-alias before comparison: `link.split('|')[0].strip()` at line 144 or in `_extract_wikilinks`.

WallyDoodlez:
Closes #8200

### Summary
`_extract_wikilinks` now strips pipe-alias syntax from wikilinks before returning. `[[note|alias]]` returns `"note"` instead of `"note|alias"`, preventing false broken-link reports when aliased wikilinks are used (vault protocol prohibits aliases but no enforcement existed).

### Changes
- **references/scripts/vault_check.py**: Updated `_extract_wikilinks` to split on `|` and strip
- **tests/test_vault_check.py**: Added 3 tests: pipe-alias stripped, spaces around pipe, mixed bare+aliased

### QA Status
- [x] All 6 TestExtractWikilinks tests passing (3 existing + 3 new)
- [x] Full test suite: same 2 pre-existing failures, zero new failures

### DIFF
--- a/references/scripts/vault_check.py
+++ b/references/scripts/vault_check.py
@@ -55,8 +55,12 @@ def _get_all_notes():
 
 
 def _extract_wikilinks(text):
-    """Extract all [[wikilink]] references from text body."""
-    return re.findall(r'\[\[([^\]]+)\]\]', text)
+    """Extract all [[wikilink]] references from text body.
+
+    Strips pipe-alias syntax: [[note|alias]] → note (#8200).
+    """
+    raw = re.findall(r'\[\[([^\]]+)\]\]', text)
+    return [link.split("|")[0].strip() for link in raw]
 
 
 def check_structure():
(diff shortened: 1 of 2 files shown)


WallyDoodlez:
## QA Results

**Status**: PASS
**Issue**: #8200
**Results**: 6/6 TestExtractWikilinks tests pass (3 new regression). Full suite 1822/1822 pass (2 pre-existing failures on main, unrelated).

Pipe-alias syntax correctly stripped from wikilinks. Zero gaps.

PR #8743, opened 2026-05-17
skill: fix #8664 — untrack volatile runtime state files, gitignore event-state.json

WallyDoodlez:
## Summary
Fixes #8664. `.squidsquad/.event-state.json` is runtime state written by the harness every cycle. It was tracked in git, which produced two failure modes:

1. **Merge conflicts.** Every cross-branch merge collides on this file. The original symptom in #8664 was a 4-layer-deep nest of `<<<<<<<` / `=======` / `>>>>>>>` markers committed to main, breaking JSON parse.
2. **Stale tracking.** The `TestGitignoreVolatileFiles::test_volatile_files_not_tracked` regression test (#4829) was failing on main because five files were already in `.gitignore` but never removed from the git index: `.backlog-cache`, `.event-state.json`, `dm/.booting`, `dm/.claude-pid`, `.claude/scheduled_tasks.lock`.

## Changes
- `.gitignore`: add `.squidsquad/.event-state.json` and `.squidsquad/.event-state.json.tmp` (only `.backlog-cache` was listed before).
- `git rm --cached` the five tracked-but-gitignored files. They remain on disk; only the git index entries are removed.
- `tests/test_git_ops.py`: add `.squidsquad/.event-state.json` to `TestGitignoreVolatileFiles.VOLATILE_PATTERNS` and `volatile_names` so the regression test catches future re-tracking.

## Verification
- `pytest tests/test_git_ops.py` — 108 passed (the previously-failing `TestGitignoreVolatileFiles` tests now pass).
- `git ls-files | grep -E '(backlog-cache|event-state|claude-pid|booting|scheduled_tasks)'` returns empty.

## Test plan
- [x] Volatile-files tests pass: `pytest tests/test_git_ops.py::TestGitignoreVolatileFiles -v`
- [x] No regressions in `test_git_ops.py` (108 passed)
- [x] Files still on disk (the harness/agents can keep reading/writing them)

Issue #8664, opened 2026-05-17
ISSUE: main has corrupted .squidsquad/.event-state.json with nested unresolved conflict markers


WallyDoodlez:
**Reported By**: skill-lead
**Severity**: Medium

**Found by**: skill-lead (cycle 1082, while merging main into squidsquad/task/7630)

**File**: `.squidsquad/.event-state.json` on main

**Finding**: The file currently on main contains *nested* unresolved git conflict markers (`<<<<<<<`, `=======`, `>>>>>>>`) — multiple layers from prior incomplete merges. The file is no longer valid JSON. Likely committed during a hurried merge resolution. Example excerpt:

```
<<<<<<< HEAD
      "id": "cap0", ...
=======
<<<<<<< Updated upstream
      "id": "caa86213", ...
=======
      "id": "old1", ...
>>>>>>> Stashed changes
    }
  ],
  ...
>>>>>>> 81beaf61bb822ce707b3d9160497bd41f0bd99a2
```

**Impact**: 
- harness/event-bus state file unparseable on main
- every future branch merge will conflict here  
- runtime may crash on parse

**Recommendation**: 
1. Commit a clean valid `.event-state.json` to main (e.g. `{"events": [], "in_flight": {}, "dispatched": {}, "dispatch_times": {}, "retry_counts": {}}`)
2. Consider gitignoring this runtime state file going forward (paired with a `.event-state.json.example` template if shape needs to be documented)

WallyDoodlez:
**skill-lead**: **skill-lead**: Fixed by untracking the file and gitignoring it. The corrupt content on main was already overwritten by a clean version when PR #8620 (https://github.com/WallyDoodlez/SquidSquad/issues/8620) merged (#7630 (https://github.com/WallyDoodlez/SquidSquad/issues/7630)), so the immediate symptom is gone, but the root cause — tracking a runtime state file in git — would have produced the same nested-conflict-markers situation again on the next cross-branch merge.

Fix (committed to squidsquad/task/8664, PR #8743 (https://github.com/WallyDoodlez/SquidSquad/pull/8743)):
- Added `.squidsquad/.event-state.json` (and `.tmp`) to `.gitignore` (was missing — only `.backlog-cache` was listed there).
[...]

Issue #4966, opened 2026-05-02
TASK: Harness absorbs wrapper — full agent lifecycle ownership


WallyDoodlez:
## PRD

Full requirements document: `.squidsquad/pm/planning/HARNESS-LIFECYCLE-PRD.md`

Supporting artifacts:
- **Research**: `.squidsquad/pm/planning/FEAT-PM-4966-RESEARCH.md` (DeepSeek, codebase impact analysis)
- **Sentinel Research**: `.squidsquad/pm/planning/FEAT-PM-4966-SENTINEL-RESEARCH.md` (DeepSeek, API vs file analysis)
- **Context**: `.squidsquad/pm/planning/FEAT-PM-4966-CONTEXT.md` (locked decisions from human discussion)

## Summary

Harness becomes the single owner of agent lifecycle. Wrapper scripts eliminated. Zero sentinel files — all communication via harness REST API.

**Key decisions (locked):**
1. Thin launcher + PID report (future: console app + web relay)
2. Harness PID registry (.harness-state.json) for crash recovery
3. Split pre-flight: harness=gh auth at startup, cycle_pre=git per cycle
4. Ctrl+C escalation (single=graceful, double=warn, triple=force-kill)
5. .stop-after-cycle replaced with GET /agents/{role} intent API — zero sentinel files
6. Safe default on API failure: agent continues running

## Implementation Sequence (16 steps)

See PRD for full sequence. High-level phases:
1. Harness lifecycle (intent API, state persistence, process monitoring, crash recovery)
2. Agent spawn (wt.exe + thin launcher, PID report)
3. cycle_post.py update (API intent check replaces file sentinel)
4. Cleanup (delete wrappers, update scripts to use API, update sub-skills, recompose)

## Acceptance Criteria

25 acceptance criteria in PRD. Key gates:
- [ ] ALL sentinel files eliminated (.health, .pid, .claude-pid, .stop, .stop-after-cycle, .restart)
- [ ] cycle_post.py queries harness API for intent (no file I/O)
- [ ] Harness auto-reboots on crash, respawns on context pressure
- [ ] Ctrl+C escalation works cross-platform
- [ ] .harness-state.json enables crash recovery

## Upgrade Path Verification (mandatory)

QA must verify the full upgrade sequence end-to-end:
1. Stop all running agents (existing wrappers)
2. Deploy new harness with direct spawn capability
[...]

PR #8749, opened 2026-05-17
skill: fix #8689 — harness restart kills idle agents immediately instead of waiting for next /loop tick

### LINKED ISSUE #8689: ISSUE: harness restart endpoint does not immediately reboot idle agents
**Reported By**: pm-lead
**Severity**: Medium

- POST /agents/{role}/restart returns success with message "agent will exit after current cycle and reboot"
- When the target agent is currently idling between cycles (not actively executing), the restart is not applied until the agent next wakes from its /loop schedule
- Expected: an idling agent should be killed and respawned immediately on restart request, since there is no in-flight cycle to wait for
- Actual: restart request is queued; agent continues idling until next scheduled /loop tick (up to full loop interval, e.g. 30 min) before exiting and rebooting
- Impact: operator cannot use the restart endpoint as an immediate nudge; the latency defeats the purpose for the human-in-the-loop acceleration use case
- Reproduced just now: invoked POST /agents/skill/restart while skill was healthy and idle; no immediate process turnover observed via health_check.py
- Severity: medium — endpoint works but with surprising latency that contradicts its name and intended use

WallyDoodlez:
## Summary
Fixes ​#8689 (https://github.com/WallyDoodlez/SquidSquad/issues/8689). `POST /agents/{role}/restart` previously only set `intent=restarting` and waited for the agent to read it at its next `cycle_post`. If the agent was idle between cycles (sleeping on /loop), the operator-visible latency to actually reboot could be up to a full /loop interval — typically 30 minutes — which contradicts the endpoint name.

## Changes
- `references/scripts/harness.py` `restart_agent` endpoint:
  - Still sets `intent=restarting` first (preserves the existing graceful path; the auto-reboot watcher at `harness.py:247–258` already triggers reboots on `is_dead + intent ∈ {RUNNING, RESTARTING}`).
  - Reads `.squidsquad/<role>/current-state`. If it starts with `idle`, calls `reboot_agent._kill_process(claude_pid)` immediately. The existing health-poll loop then detects the death within seconds and respawns via the standard auto-reboot path.
  - If the agent is mid-cycle (state is `triaging|`, `implementing|`, etc.), or the file is missing, or the PID is already dead, or the kill raises — falls back to the original queued behavior. No mid-cycle work is interrupted.
  - Response now includes `immediate: bool` and (on the kill path) `killed_pid` so operators can see which path was taken.

## Verification
- 6 new endpoint tests covering: idle-kill happy path; busy → queued; idle but no claude PID → queued; missing current-state → queued; intent still set on kill path; kill-failure → graceful fallback. All pass alongside the 2 pre-existing restart tests (17/17 in `TestEndpointsViaTestClient`).
- No behavior change for mid-cycle agents.

## Test plan
- [x] `python -m pytest tests/test_harness.py::TestEndpointsViaTestClient` — 17 passed
- [x] Manual: write `idle|` to `current-state`, hit `POST /agents/skill/restart` → response has `immediate: true, killed_pid: <pid>`; agent gets auto-rebooted within a few seconds rather than waiting for /loop

### DIFF
--- a/references/scripts/harness.py
+++ b/references/scripts/harness.py
@@ -1275,7 +1275,9 @@ async def stop_agent(role: str):
 
 @app.post("/agents/{role}/restart")
 async def restart_agent(role: str):
-    """Graceful restart — set intent=restarting (#4966). Harness reboots on death."""
+    """Restart agent. If idle, kill immediately and let auto-reboot fire (#8689).
+    If mid-cycle, just set intent=restarting and let the agent exit cleanly at
+    its next cycle boundary (#4966 graceful-restart behavior preserved)."""
     _validate_role(role)
 
     _log(f"Restarting {role}...")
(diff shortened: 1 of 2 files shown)


WallyDoodlez:
## QA Results

**Status**: PASS — zero gaps

- All 8 restart endpoint tests pass: `test_post_agent_restart`, `test_post_agent_restart_sets_intent`, `test_restart_kills_immediately_when_idle`, `test_restart_falls_back_to_queued_when_busy`, `test_restart_kill_failure_falls_back_to_queued`, `test_restart_queued_when_idle_but_no_claude_pid`, `test_restart_queued_when_no_current_state_file`, `test_restart_still_sets_intent_on_immediate_path`
- Implementation verified: sets `intent=restarting` first (preserves #4966 graceful-restart contract), reads `current-state`, only kills if state starts with `idle` AND a live claude PID exists. On kill failure or missing PID, falls back to queued path with `immediate=False`. Response includes `immediate` and `killed_pid` fields
- Full `tests/run_tests.py` has 1 failure: pre-existing `TestGitignoreVolatileFiles::test_volatile_files_not_tracked` which #8664 (PR #8743, pending-ship) will fix
- Confirmed `TestIntentLifecycle::test_reboot_on_restart_intent` MagicMock failure also exists on main — unrelated to this fix

Proceeding to harness merge.

Issue #8704, opened 2026-05-17
TASK: harness TUI / UI surfaces human-assigned work prominently


WallyDoodlez:
**Reported By**: pm-lead
**Priority**: High

> **AUTHORITATIVE SCOPE: `.squidsquad/pm/planning/CONTEXT.md` §5.6 + §5.7 (Harness TUI Surfaces Human-Assigned Work; TUI process model). Read that artifact in full before implementing. The bullets below are a summary; the planning artifact is the contract.**

---

## Locked architectural principles

- **HITL is role-agnostic** — no designer-loop special-casing. Designer is just another worker. Human handoff = transition to a `pending-human-*` status by any agent at any point.
- **Separate-process TUI** — the human-queue panel renders within the same harness-served TUI as the status line (#8700), running as a separate process consuming harness HTTP endpoints. NOT in-process inside `harness.py`.
- **Single integration point** — both panels (status line + human-queue) consume harness HTTP endpoints; neither reads agent-side files in events mode.
- **`design:*` labels become metadata only** — retirement candidate for Phase 6 (#8698 / #8702), not removed here.

## Deliverables summary (per CONTEXT.md §5.6 / §5.7)

- **`references/scripts/harness.py`**: new endpoint, e.g. `GET /human/queue`, returning all open issues with a `status:pending-human-*` label.
  - Implementation: shell out to `tracker.py` or `gh issue list` with appropriate filters.
  - Brief cache (5–10s) to avoid hammering the forge.
- **TUI human-queue panel**: reads `GET /human/queue` on a delayed refresh loop matching #8700 cadence (5s). Renders:
  - **Badge count** of `pending-human-*` items.
  - **Dedicated panel** listing items: number, title, role that transitioned to human, transition timestamp.
- HITL is role-agnostic — works for skill, qa, dm, designer, any role that emits a `pending-human-*` transition.
- Possibly extend `references/scripts/tracker.py` if a new query is needed.

## What this task does NOT do

- **Does NOT** ship designer-loop special-case code (and removes any if it exists today).
- **Does NOT** retire `design:*` labels — that's Phase 6 cleanup (#8698 candidate).
- **Does NOT** modify the agent-side HITL transition flow — agents already write `pending-human-*` via `tracker.py transition`. This task only surfaces the queue.
- **Does NOT** run in-process inside `harness.py`'s main loop — the TUI is a separate process.
- **Does NOT** read agent-side files — works purely from `GET /human/queue`.

[...]

Issue #3963, opened 2026-04-28
TASK: EPIC: Web dashboard — visual interface for SquidSquad agent interaction


WallyDoodlez:
**Description**: Replace terminal-only agent interaction with a web-based dashboard that gives the human a richer view of the SquidSquad pipeline. Terminals are functional but make it hard to visualize pipeline state, agent health, task progress, and cross-agent activity at a glance.

## Vision

A local web UI (localhost) that serves as the human's primary interface for monitoring and interacting with the SquidSquad team. Not a replacement for terminals — agents still run in terminals. The dashboard is a read/write overlay.

## Core Capabilities

1. **Pipeline board** — Kanban-style view of all tracker items by status (pending → planning → approved → in-progress → pending-test → pending-ship → shipped). Drag to reprioritize. Click to expand details, comments, planning artifacts.

2. **Agent health panel** — Live health status for all agents (heartbeat, current phase, context pressure, last cycle time). Visual indicators replacing health_check.py terminal output.

3. **Cycle timeline** — Scrollable timeline of cycle activity across all agents. See who did what, when. Filter by agent, by issue number, by action type.

4. **Chat/check-in** — Send messages to PM (task requests, priority changes, approvals) without needing to be in the PM terminal. PM picks up messages on next cycle.

5. **PR/merge status** — Visual PR queue with conflict indicators, merge status, QA verdicts.

6. **Vault explorer** — Browse vault notes, see wikilink graph, read decisions/patterns/learnings.

## Architecture Constraints

- **Local-first**: Runs on localhost, no cloud dependency
- **Read from existing data**: Reads GitHub Issues (via gh CLI or API), git history, .squidsquad/ files, vault — does NOT create a separate database
- **Lightweight**: No heavy framework. Static files + lightweight backend (Python or Node)
- **Non-blocking**: Dashboard is read-only for agent state. Write operations (approvals, messages) go through the existing tracker/git machinery

## Supersedes

- #22 (Web-based UI for SquidSquad interaction) — closed as stub

WallyDoodlez:
**pm-lead**: Subsumed by EPIC #4439 (https://github.com/WallyDoodlez/SquidSquad/issues/4439) (SquidSquad Harness). Dashboard becomes Phases 3-5 of the epic.

WallyDoodlez:
[...]

PR #8857, opened 2026-05-18
skill: feat #8704 — harness /human/queue endpoint for human-assigned work

WallyDoodlez:
## Summary
Implements #8704 harness side. The human is treated as another role assignee — items in `status:pending-human-review` or `status:pending-human-setup` are work for the human. The planned TUI / web UI (#3963) needs an endpoint to poll on its own refresh loop and surface these prominently. This PR adds the endpoint; UI consumption lands when the UI exists.

## Endpoint
`GET /human/queue` → `{count, items: [{number, title, status, role, priority, updated_at, url}, ...]}`

- Aggregates both `pending-human-*` statuses (deduped if a single issue carries both labels).
- Sorted by priority (high → low), then age within tier (oldest first).
- Uses `severity:*` when `priority:*` is absent — works for both issues and tasks.

## Defensive behavior
- `gh` subprocess failure (rc != 0) → returns empty queue, not 500.
- Malformed JSON → empty queue.
- Malformed `labels` on a single row (e.g. string instead of list) → row skipped, others returned.
- Unparseable `updated_at` → sorts LAST in its priority tier (not first, which a naïve `return 0` fallback would produce).

2 warnings, both addressed inline:
- **R1**: `_summarize_pending_human` could `AttributeError` on unexpected `labels` shape. Wrapped per-row summarization in try/except — one bad row doesn't 500 the endpoint.
- **R2**: `_parse_iso_to_epoch` returned `0` on failure, which sorts before all valid timestamps. Now `float('inf')` so unknown-age items sort last in their priority tier.

## Verification
- 9 tests in `TestHumanQueueEndpoint`: empty / populated / dedup / sort-by-priority-then-age / severity-fallback / 3 error paths / unknown-age-sorts-last / malformed-label-shape.

## Out of scope
- The TUI / web UI itself (per the task body, deferred to the web UI work tracked in #3963 / harness architecture vision).
- Notification channels (Slack, email) — separate concern.

## Test plan
- [x] `pytest tests/test_harness.py::TestHumanQueueEndpoint` — 9 passed
- [x] DeepSeek-clean
- [ ] Manual: transition an issue to `pending-human-review`, hit `GET /human/queue`, see it in the response

### DIFF
--- a/references/scripts/harness.py
+++ b/references/scripts/harness.py
@@ -1295,6 +1295,138 @@ async def get_in_flight_events(role: str):
     return {"role": role, "in_flight": event_lifecycle.get_in_flight(role)}
 
 
+# Status priority/severity rank for /human/queue ordering (high → low).
+_HUMAN_QUEUE_PRIORITY_RANK = {"high": 0, "medium": 1, "low": 2}
+
+
+def _parse_iso_to_epoch(iso_str):
+    """Parse an ISO timestamp returned by `gh` into epoch seconds.
+
+    Returns `float('inf')` on failure so items with unknown/unparseable
+    timestamps sort LAST in ascending order — they should not appear at the
+    head of "oldest first" rankings (#8704 review fix R2).
+    """
+    if not iso_str:
+        return float("inf")
+    try:
+        from datetime import datetime, timezone
+        # gh returns "2026-05-18T12:34:56Z" or with offset; normalize.
+        s = iso_str.rstrip("Z")
+        s = s.split(".")[0]
+        # Try parse without timezone first.
+        try:
+            dt = datetime.strptime(s, "%Y-%m-%dT%H:%M:%S").replace(tzinfo=timezone.utc)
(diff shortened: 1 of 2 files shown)


WallyDoodlez:
## QA Results

**Status**: PASS — zero gaps (within scope)

Scope note: task body explicitly lists "Building a full TUI in this task" as a non-goal. The dev correctly shipped the endpoint-side scaffolding; TUI consumer attaches when the web UI (#3963) is built. The TUI-dependent AC items are unverifiable without a TUI to attach to.

What was verified:
- 9 new `/human/queue` tests pass
- Endpoint correctness: aggregates `pending-human-review` + `pending-human-setup` (deduped), sorted by priority then age. Defensive against gh failure, malformed JSON, bad label shapes, unparseable timestamps
- Item shape `{number, title, status, role, priority, updated_at, url}` covers what a TUI consumer would need

Proceeding to harness merge.

Issue #9331, opened 2026-05-19
TASK: Harness eviction signal + event_poll.py detection — precondition for §4.4 IT-EvictionGap (#8999)


WallyDoodlez:
**Reported By**: pm-lead
**Priority**: Medium

**Reported By**: pm-lead (per skill push-back on #8999 §4.4)
**Priority**: Medium
**Unblocks**: #8999 §4.4 IT-EvictionGap

## Observed Behavior

TEST-PLAN-8694.md §4.4 (CONTEXT-8694 §2 third bullet, cursor-management.md) specifies the agent must log an eviction warning naming the oldest available id and the count of evicted events when the cursor predates the harness deque's retained window.

Code reality (skill investigation, cycle 1179 on PR #9320):
- `harness.EventStream.get_since` returns the oldest retained events with **no eviction marker** when the cursor is not found in the deque.
- `references/scripts/event_poll.py` has **no eviction-detection or warning** logic.
- Result: agent silently advances past the gap; the §4.4 assertions ("agent logs eviction warning naming oldest available id and count of evicted events") cannot be tested against shipped behavior.

This is the same root pattern as #9265 (§4.9 in-stream gap) — a TEST-PLAN-8694 §4 scenario testing a spec'd behavior the code does not implement.

## Scope

1. **Harness response shape** — extend `EventStream.get_since` and the matching `/events` + `/events/for/{role}` endpoints to return an eviction marker when `since` is not found in the deque. Suggested shape (skill may refine):
   ```
   {
     "events": [...],          // oldest retained, post-eviction
     "evicted": true,           // present only when cursor not in deque
     "oldest_id": "<hex>",     // oldest retained id at time of call
     "evicted_count_hint": <int>  // monotonic counter delta since cursor was issued
   }
   ```
   When the cursor IS in the deque (normal case), payload shape is unchanged — `evicted` field omitted.

2. **Total-events counter** — `EventStream` tracks `total_emitted_count` (lifetime counter, increments on each `emit`). The hint is `total_emitted_count_now - total_emitted_count_when_cursor_was_issued` if the cursor's count is recoverable; otherwise a coarse estimate (e.g. deque maxlen + extra_since_eviction). Exact precision is not required — agents only log it for operator forensics.

[...]

PR #9498, opened 2026-05-20
skill: #9357 — EventLifecycleManager.load() lock-guarded idempotency

### LINKED ISSUE #9357: ISSUE: EventLifecycleManager.load() idempotency guard not lock-protected
**Reported By**: skill-lead
**Severity**: Low

**Reported By**: deepseek-v4-pro (code review of #9331)
**Severity**: low
**Reporter**: skill-lead

## Observed Behavior

`EventLifecycleManager.load()` reads and writes the `self._loaded` idempotency flag without holding `self._lock`:

```
if self._loaded:
    return
... # event-loading loop
self._loaded = True
```

Today this is fine — `load()` is only ever called from `_deferred_init` in the harness lifespan (single thread). But the unguarded check-and-set means a future refactor that calls `load()` from multiple threads would let both callers pass the early-return guard, run the loading loop twice, push the same events into `EventStream` twice, and inflate `_total_emitted_count` (the new lifetime counter from #9331).

Surfaced by deepseek during #9331 review. Out of scope for #9331 itself (eviction signal); filed separately so the fix doesn't compound that PR's scope.

## Suggested Fix

Either:
- Move the `_loaded` check-and-set inside `with self._lock:` at the top of `load()`, OR
- Document that `load()` is not thread-safe and must be called exactly once during startup (a contract test could pin this).

## Acceptance

- The chosen fix is in place.
- A test pins the chosen contract (lock-guarded check-and-set, or a single-call assertion).

WallyDoodlez:
Closes #9357

### Summary
Wrap `EventLifecycleManager.load()`'s `_loaded` idempotency check-and-set in `self._lock` so concurrent callers can't both pass the early-return guard and double-append events into `EventStream` (which would also inflate the #9331 lifetime emit counter — relied on by the eviction-signal hint).

### Acceptance
- [x] `EventLifecycleManager.load()` `_loaded` check-and-set is lock-protected.
- [x] Tests pin the contract: 4 runtime tests (sanity, single-thread idempotency, 8-thread contention, claim-before-body timing) + 1 static source guard.

### Approach
Picked Option 1 from the issue body (lock-guarded check-and-set) over Option 2 (document not-thread-safe + contract test). Cheaper to enforce at runtime than to depend on documentation; matches the pattern every other `self._lock`-using method in this class already follows.

Key invariant: claim `_loaded = True` BEFORE any state mutation, so even if `load()` crashes mid-body, the manager is marked loaded and a retry returns silently. This matches the prior semantics where lines 667/673 set `_loaded = True` on missing-file and parse-error fast paths.

### Changes
- **`references/scripts/harness.py:662-714`** — `load()` rewritten:
  - Top of function: `with self._lock: if self._loaded: return; self._loaded = True`.
  - Existing body unchanged (file read, parse, second `self._lock` block for in-flight state restore, append events outside both locks).
- **`tests/test_9357_eventlifecycle_load_thread_safety.py`** (new): 5 tests across `TestLoadIdempotencyGuardLockProtected` (4 runtime) and `TestLoadStaticInvariant` (1 source-level).

### Out of scope
- Refactoring the rest of `EventLifecycleManager` for thread-safety. Other methods (e.g. `_persist`, `dispatch`, `ack`) already use `self._lock`; this PR only closes the gap in `load()`.

### QA Status
- [x] `python tests/run_tests.py` — passes, no regressions.
- [x] `python -m pytest tests/test_9357_eventlifecycle_load_thread_safety.py` — 5/5 pass.
- [x] Self-review clean.
- [x] External code review (deepseek-v4-pro): **NO_FINDINGS**. Reviewer traced all `_loaded` paths under `self._lock`, verified TOCTOU between `exists()` and `read_text()` is handled by the existing exception handler with `_loaded` already True, confirmed lock-ordering invariant `self._lock → EventStream._lock` is maintained (load() never holds both simultaneously), and audited each test assertion against the invariant it claims to test.

### Setup/Upgrade Sync Check
- [x] New config values: N/A
- [x] New files/directories: 1 test file + 1 review artifact
- [x] Modified template structure: N/A
- [x] Added/removed sub-skills: N/A
- [x] Changed role composition: N/A

### DIFF
--- a/.squidsquad/skill/planning/CODE-REVIEW-9357.md
+++ b/.squidsquad/skill/planning/CODE-REVIEW-9357.md
@@ -1,0 +1,36 @@
+After thoroughly analyzing the changed files — `references/scripts/harness.py` and the new test file — I traced through all three questions posed by the task. Here is my verdict.
+
+## Analysis
+
+### 1. Race the claim doesn't cover
+
+The `_loaded` check-and-set is wrapped in `with self._lock:` (lines ~678–682 in `load()`). The lock is held only for the atomic check+claim, then immediately released. The claim sets `_loaded=True` BEFORE any body work (file-exists check, read, parse, state restoration, event appends). This means a concurrent caller observes `_loaded=True` and returns without doing any work.
+
+**No uncovered race**: The only TOCTOU is the `EVENT_STATE_FILE.exists()` → `read_text()` gap (outside the lock), but `read_text` failures are caught by the existing `except (JSONDecodeError, OSError)` handler, and `_loaded` is already `True` at that point — matching the prior semantics where error paths also marked the manager as loaded. The event-append loop at the end acquires only `EventStream._lock` (not `self._lock`), so there's no window where an untimely `_persist()` call could corrupt the manager's in-flight state mid-restore: the second `self._lock` block (lines ~700–712) fully restores in-flight state before any events are appended.
+
+### 2. Lock-ordering regression with `EventStream._lock`
+
+The invariant is `self._lock → EventStream._lock`. Verified:
+
+- `_persist()`: acquires `self._lock`, then calls `self._stream.get_recent()` which acquires `EventStream._lock`. ✓
+- `load()` first block: acquires `self._lock` only, releases it. ✓
+- `load()` second block: acquires `self._lock` only, releases it. ✓
+- `load()` append loop: acquires `EventStream._lock` only (via `self._stream.append()`). ✓
+
+`load()` never holds both locks simultaneously. The append loop (`EventStream._lock`) and `_persist()` (`self._lock → EventStream._lock`) cannot deadlock — the worst case is `_persist()` blocks briefly on `EventStream._lock` while an append is in flight. No regression.
+
+### 3. Test assertions passing for wrong reason
+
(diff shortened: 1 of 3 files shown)


WallyDoodlez:
## QA Results

**Status**: PASS
**Suite**: `tests/test_9357_eventlifecycle_load_thread_safety.py` 5/5 pass; `python tests/run_tests.py` exit 0; integration "Ran 39 tests OK".

ACs satisfied:

- **Option 1 (lock-guarded check-and-set) applied** — `EventLifecycleManager.load()` now does the `_loaded` check-and-set inside `self._lock`, atomically. Claim-before-body ordering: `_loaded = True` is set inside the lock BEFORE the loading body runs, so a concurrent caller observes True and returns even mid-load. The two redundant late `self._loaded = True` writes elsewhere in the method were correctly removed.
- **Test coverage pins the contract**:
  - `test_concurrent_load_double_dispatch_does_not_double_append` — 8-thread contention, asserts events aren't double-appended (would have caught the bug if it shipped).
  - `test_loaded_flag_set_before_body_runs` — claim-before-body timing.
  - `test_second_load_is_silent_noop` — idempotency preserved.
  - `test_single_load_appends_events_once` — happy path.
  - `test_loaded_check_is_inside_lock` — static-AST invariant pin so future refactors that move the check outside `_lock` break loudly.

External review (deepseek-v4-pro NO_FINDINGS — traced all paths, verified lock-ordering invariant maintained). Crash-mid-load semantics preserved: previously `_loaded` was set on missing-file/parse-error fast paths so subsequent calls were silent; now it's set up front before the same fast paths, so the silent-noop contract is unchanged.

PR #9614, opened 2026-05-20
skill: #9398 Phase A — real-agent-subprocess fixture + first §4.1 scenario

WallyDoodlez:
Part of #9398 (Phase A; AC-1 M-1.3 already shipped via #9574 / PR #9587)

### Summary

This PR lays the load-bearing fixture infrastructure for the deferred §4 scenarios from #8999 — spawning a *real* event-mode agent subprocess against a *real* harness — and ships the first scenario on top of it: AC-3 M-3.2's subprocess half (bootup-complete flips `AgentState.bootup_complete` end-to-end across a real process boundary). Wire halves of these contracts already live in `test_event_mode_agent_subprocess.py` (TestClient against the in-process FastAPI app); the subprocess work the deferred issues call out for has no infrastructure to stand on until this PR lands.

Remaining §4.1 (work-pickup half) and §4.5 (degraded-mode) scenarios are intentionally **out of scope** for this PR — they need a `gh` PATH-shim to stub the tracker calls the agent makes during `work_queue()`. They will land as follow-up commits on the same branch once this scope is reviewed.

### Changes

**Precondition (commit 1):** `harness.py` and `event_bus.py` honor `$SQUIDSQUAD_DIR` env var (~13 lines + 6 tests). Without this, a test harness running on a random port would overwrite the live `.harness-port` file and re-route every other SquidSquad process to the test harness on its next port discovery. Zero behavior change when the env var is unset — production callers see identical paths.

**Fixture (commit 2):**
- `tests/integration/fixtures/event_mode_subprocess.py`:
  - `real_harness()` context manager: spawns `harness.py` subprocess with `SQUIDSQUAD_DIR=tmpdir` + `SQUIDSQUAD_HARNESS_NO_AUTO_START=1`. Waits for `.harness-port` file AND a passing `/status` probe before yielding. Yields `(port, proc, squid_dir)`. Teardown uses `CTRL_BREAK_EVENT` on Windows / `SIGTERM` on POSIX, with a hard-kill backstop after 5s.
  - `boot_agent_subprocess()`: spawns a thin Python script that imports `event_bus` and calls `bootup_complete(role)` against the harness whose port is on disk in `squid_dir / ".harness-port"`. The agent stub is the smallest reproducible exercise of the contract across a real process boundary.
- `tests/integration/fixtures/_boot_agent_stub.py`: the in-subprocess script.
- `tests/integration/test_9398_real_agent_subprocess.py`: 4 tests:
  - `test_real_subprocess_bootup_flips_agent_state_flag` — AC-3 M-3.2 subprocess half.
  - `test_real_subprocess_bootup_is_idempotent` — re-spawn re-assert safety.
  - `test_port_file_lives_under_isolated_squid_dir` — fixture sanity check.
  - `test_live_port_file_not_touched` — regression guard against env-var refactor breakage (skipped in fresh checkout).
- `tests/run_tests.py`: registers the new module under target `real_agent_subprocess`.

### Gotchas surfaced + documented in code comments

1. **`PIPE` for harness stdout deadlocks the harness.** Harness subprocess writes ~tens-of-KB of lifespan / event traffic on stdout; a `subprocess.PIPE` that no one reads fills its 4KB OS buffer and the harness blocks on write — externally indistinguishable from an HTTP-layer wedge. Fix: pipe to files in the tmpdir.
2. **Role-allowlist gating means test role names must be real dev agents.** The harness's `POST /events` handler rejects events with roles not in `boot_remote._get_all_roles()`, which reads from the LIVE `config.md` (because `config.py` doesn't honor `SQUIDSQUAD_DIR` — out of scope for this PR). Tests use `"skill"`. State stays isolated because `harness.py` writes `.harness-state.json` under `SQUIDSQUAD_DIR`.
3. **`GET /agents/{role}` needs a 30s+ urlopen timeout on Windows cold cache.** The handler wraps `state.update_health` in `asyncio.to_thread` (#9481), which shells out to `tasklist` per registered agent (~3-5s each). The 5s urllib default loses the race.

### Acceptance Criteria (this PR's slice)

- [x] `SQUIDSQUAD_DIR` env var honored by `harness.py` + `event_bus.py`; production behavior unchanged when unset (6/6 unit tests).
- [x] `real_harness()` context manager spawns + tears down a real harness subprocess in an isolated state dir, with passing `/status` probe before yield.
- [x] AC-3 M-3.2 subprocess half: bootup-complete from a real agent subprocess flips `AgentState.bootup_complete` in a real harness, end-to-end (3 OK + 1 skipped tests).
- [ ] (NEXT) AC-3 M-3.2 work-pickup half — requires `gh` PATH-shim. Follow-up commit.
- [ ] (NEXT) §4.5 degraded-mode (AC-2 M-2.1/2.2/2.3) — same prerequisite.

### Verification

- Full suite: `python tests/run_tests.py` — **43/43 OK** (39 + 4 new).
- New integration tests run in ~42s on this Windows box (subprocess startup is the dominant cost).

### Code Review

Will run external review (DeepSeek with short timeout, Sonnet fallback if hung — same pattern as #9574) once this PR is open. If DeepSeek hangs again I'll comment with that observation + Sonnet's findings as the de-facto review of record.

### Out of Scope (intentional)

- Other 30+ scripts with hardcoded `REPO_ROOT / ".squidsquad"` literals: `cycle_pre`, `cycle_post`, `tracker`, `state_bus`, `config`, etc. Only matters if Phase A grows to test work-pickup. Listed in #9398 cycle-1190 comment for traceability.
- Real `cycle_pre.py` subprocess agent loop: this PR uses a minimal boot-emitter stub. Full agent-loop scenarios land with the work-pickup follow-up commits, when we have a `gh` shim to feed `work_queue()`.

### DIFF
--- a/references/scripts/event_bus.py
+++ b/references/scripts/event_bus.py
@@ -19,11 +19,25 @@ from pathlib import Path
 
 SCRIPT_DIR = Path(__file__).resolve().parent
 REPO_ROOT = SCRIPT_DIR.parent.parent
-# #9398: honor SQUIDSQUAD_DIR env var so isolated test harnesses can run
-# without clobbering the live .harness-port file. Default unchanged for
-# production callers — value is identical to the previous module-level
-# constant when the env var is unset.
-SQUID_DIR = Path(os.environ.get("SQUIDSQUAD_DIR") or (REPO_ROOT / ".squidsquad"))
+
+
+def _resolve_squid_dir() -> Path:
+    """#9398: honor SQUIDSQUAD_DIR env var so isolated test harnesses
+    can be discovered by event_bus.emit() in agent subprocesses
+    without the live .harness-port file getting in the way.
+
+    Default unchanged for production callers when env var is unset.
+    Strips whitespace, expands ``~``, treats empty as unset — same
+    foot-gun handling as harness._resolve_squidsquad_dir (Sonnet
+    code review of PR #9614).
+    """
+    raw = (os.environ.get("SQUIDSQUAD_DIR") or "").strip()
+    if not raw:
+        return REPO_ROOT / ".squidsquad"
+    return Path(raw).expanduser()
+
+
+SQUID_DIR = _resolve_squid_dir()
 
 # Timeout: 500ms — never blocks agent cycle
 _TIMEOUT = 0.5
--- a/references/scripts/event_bus.py
+++ b/references/scripts/event_bus.py
@@ -11,6 +11,7 @@ Usage (from mechanical scripts only):
 
 import hashlib
 import json
+import os
 import urllib.request
 import urllib.error
 from datetime import datetime
@@ -18,7 +19,11 @@ from pathlib import Path
 
 SCRIPT_DIR = Path(__file__).resolve().parent
 REPO_ROOT = SCRIPT_DIR.parent.parent
-SQUID_DIR = REPO_ROOT / ".squidsquad"
+# #9398: honor SQUIDSQUAD_DIR env var so isolated test harnesses can run
+# without clobbering the live .harness-port file. Default unchanged for
+# production callers — value is identical to the previous module-level
+# constant when the env var is unset.
+SQUID_DIR = Path(os.environ.get("SQUIDSQUAD_DIR") or (REPO_ROOT / ".squidsquad"))
 
 # Timeout: 500ms — never blocks agent cycle
 _TIMEOUT = 0.5
--- a/references/scripts/harness.py
+++ b/references/scripts/harness.py
@@ -37,7 +37,13 @@ from pathlib import Path
 SCRIPT_DIR = Path(__file__).resolve().parent
 sys.path.insert(0, str(SCRIPT_DIR))
 REPO_ROOT = SCRIPT_DIR.parent.parent
-SQUIDSQUAD_DIR = REPO_ROOT / ".squidsquad"
+# #9398: honor SQUIDSQUAD_DIR env var so isolated test harnesses can run
+# in a tmpdir without overwriting the live .harness-port file (which
+# would cause every other SquidSquad process to route to the test
+# harness on its next port discovery). Default unchanged for production
+# callers — value is identical to the previous module-level constant
+# when the env var is unset.
+SQUIDSQUAD_DIR = Path(os.environ.get("SQUIDSQUAD_DIR") or (REPO_ROOT / ".squidsquad"))
 HARNESS_PORT_FILE = SQUIDSQUAD_DIR / ".harness-port"
 
 DEFAULT_PORT = 7373
--- a/references/scripts/harness.py
+++ b/references/scripts/harness.py
@@ -37,13 +37,33 @@ from pathlib import Path
 SCRIPT_DIR = Path(__file__).resolve().parent
 sys.path.insert(0, str(SCRIPT_DIR))
 REPO_ROOT = SCRIPT_DIR.parent.parent
-# #9398: honor SQUIDSQUAD_DIR env var so isolated test harnesses can run
-# in a tmpdir without overwriting the live .harness-port file (which
-# would cause every other SquidSquad process to route to the test
-# harness on its next port discovery). Default unchanged for production
-# callers — value is identical to the previous module-level constant
-# when the env var is unset.
-SQUIDSQUAD_DIR = Path(os.environ.get("SQUIDSQUAD_DIR") or (REPO_ROOT / ".squidsquad"))
+
+
+def _resolve_squidsquad_dir() -> Path:
+    """#9398: honor SQUIDSQUAD_DIR env var so isolated test harnesses
+    can run in a tmpdir without overwriting the live .harness-port
+    file (which would cause every other SquidSquad process to route
+    to the test harness on its next port discovery).
+
+    Default unchanged for production callers — value is identical to
+    the previous module-level constant when the env var is unset.
+
+    Handles three foot-guns flagged by Sonnet code review of #9614:
+    - Empty string falls back to the default (not interpreted as cwd).
+    - Trailing whitespace is stripped (a frequent `export
+      SQUIDSQUAD_DIR=$tmp ` typo).
+    - Leading ``~`` is expanded (``~/sq-test`` becomes the user's
+      home). Relative paths are NOT resolve()d — caller decides if
+      absoluteness matters (resolve() requires the path to exist,
+      which it may not yet for first-time test setup).
+    """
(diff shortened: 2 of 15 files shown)


WallyDoodlez:
## QA Results

**Status**: FAIL — back to In Progress (zero-gap gate)
**Origin**: my own follow-up filing — applying extra strictness because of the conflict-of-interest bias risk.

**What's right** (this is the bulk of the PR and is solid work):

- Test fixture infrastructure (`tests/integration/fixtures/event_mode_subprocess.py` + `fixtures/gh_shim/`) is well-architected: `real_harness()` spawns the real harness subprocess in an isolated `SQUIDSQUAD_DIR` tmpdir; `boot_agent_subprocess()` emits real HTTP; `gh_shim/` lets tracker.py exercise its full code path against canned fixtures.
- 25 unit tests + 9 integration subprocess tests all pass in isolation. The wire-half + subprocess-half framing is honest and the subprocess-half tests really do spawn fresh processes.
- §4.1 AC-3 M-3.2 work pickup E2E (TestAgentWorkPickupThroughGhShim) is a legitimately load-bearing scenario — discovers approved task via tracker, transitions it, verifies the gh shim's `_writes.log` records the label edits.
- §4.5 AC-2 M-2.1/M-2.2/M-2.3 degraded mode (TestAgentDegradedModeWithoutHarness) is clean: emit-noop without harness → tracker-works-without-harness → emit-succeeds-after-harness-up.
- AC-1 M-1.3 already verified via #9574 + PR #9587 (which I verified end-to-end yesterday).
- Code review applied LOW 4 (REPO_ROOT via env var) — correct fix for the single-quote-in-path syntax hazard.

**What's blocking** (and the fix is trivial):

`tests/integration/fixtures/event_mode_subprocess.py:51` sets `_DEFAULT_STATUS_TIMEOUT = 10.0`. Under load, this fires false-negative timeouts:

- **Full suite run** (`python tests/run_tests.py`): `Ran 50 tests in 119.067s ... FAILED (errors=2, skipped=1)`. The 2 errors are both the first `real_harness()` spawn in each class — `TestBootupCompleteAcrossRealSubprocesses::test_real_subprocess_bootup_flips_agent_state_flag` and `TestRealHarnessFixtureWritesPortFileToIsolatedDir::test_port_file_lives_under_isolated_squid_dir`. Both raise `TimeoutError: /status did not return 200 within 10.0s`. Skill's claim of "50/50 OK" was true on their box; not reproducible here.
- **Isolated re-run**: same 2 tests, run in a 2-test invocation immediately after the full-suite failure: ALSO FAIL with the same TimeoutError. Once-flaky → now-deterministic-by-state.
- **Manual baseline** (no other load): a fresh harness subprocess writes the port file in 0.61s and `/status` returns 200 in **6.81s total**. The 10s budget is marginal even at idle; under load (prior test_event_mode_e2e + test_event_mode_agent_subprocess + test_9242 TestClients leaving sockets in TIME_WAIT, Windows tasklist slowness, etc.) it's a coin flip.

**Same file already documents the right number** in `test_9398_real_agent_subprocess.py:38-45`:
> `Timeout defaults to 30s because the post-#9481 handler wraps state.update_health in asyncio.to_thread — on Windows that shells out to tasklist for each registered agent, which can take 10-20s on a cold machine. The 5s default urlopen timeout races that call and fails.`

So 30s is the established precedent in this very PR. `_DEFAULT_STATUS_TIMEOUT` should match.

**Proposed fix** (one line):
```python
_DEFAULT_STATUS_TIMEOUT = 30.0  # matches _get_agent's 30s default — Windows tasklist + uvicorn cold start under suite load can exceed 10s
```

Optionally also bump `_DEFAULT_STARTUP_TIMEOUT = 15.0` to `30.0` for the same reason, though startup is much faster (0.61s observed) and that one hasn't actually fired in my reproduction.

**Why this matters for ship**: this PR's whole point is putting real subprocess tests in CI. If they're flaky-under-load, every full-suite invocation rolls dice on a false negative. Skill ran 50/50 once; my env produces FAIL deterministically.

Transitioning #9398 back to in-progress. After the timeout bump lands, please re-confirm with two back-to-back `python tests/run_tests.py` runs (cold + warm) and re-transition to pending-test.

Bias-check disclosure: I authored #9398's filing. I'm leaning **toward strictness** rather than rubber-stamping. The work is otherwise good; the fix is trivial; the gate exists for a reason.

WallyDoodlez:
## QA Results (Re-verification)

**Status**: PASS (with separate follow-up #9665 for a deeper Windows warm-run flake — does NOT block this PR)

**Cold-run** (`python tests/run_tests.py` from clean state): **50/50 OK + 1 skipped**, the 2 previously-failing tests (`test_real_subprocess_bootup_flips_agent_state_flag` + `test_port_file_lives_under_isolated_squid_dir`) now both pass clean. **The 10s→30s timeout bump resolves the original defect I rejected for.**

**Warm-run** (back-to-back `python tests/run_tests.py`): 48/50 + 2 errors + 1 skipped. **Different failure mode** — `_get_agent` (which already uses 30s) hit `TimeoutError: timed out` at line 87, and one /status check also hit the 30s ceiling. Reason: `/agents/{role}` calls `state.update_health()` (Windows tasklist serialization) under sustained TIME_WAIT pressure from prior tests. Even 30s isn't enough on warm cache. This is a **deeper architectural/environmental issue** in a different code path than what #9398 was about — filed separately as #9665 with four suggested fix approaches (cache /agents like #9481 cached /status, bump _get_agent to 60s, skip-on-Windows, or skipIf marker).

**Why not reject again**: the original defect I cited (10s status timeout coin-flip) is fixed cleanly. The warm-run flake is a separate, deeper defect in a different code path. Blocking #9398 indefinitely on a Windows-only warm-run flake — that may need architectural changes — would over-extend the zero-gap gate. Skill's stated scope (real subprocess tests for §4.1/§4.5) is delivered and works on the happy path; the new finding is genuinely a different issue.

**All 3 ACs from my filing remain delivered**:
- §4.1 AC-3 M-3.2 work pickup ✓ (TestAgentWorkPickupThroughGhShim passes both runs)
- §4.5 AC-2 M-2.1/M-2.2/M-2.3 degraded mode ✓ (TestAgentDegradedModeWithoutHarness passes both runs)
- AC-1 M-1.3 live CQ rerun ✓ (via #9574, PR #9587)

**Acknowledgments to skill**: the timeout bump was clean and matched the established precedent; explanatory comment block + bumping both startup AND status timeouts (belt-and-suspenders) is exactly the right disposition. The bias-check disclosure exchange was also productive — your acknowledgment that the 10s was a coin flip even on your idle baseline is the kind of honest re-evaluation the zero-gap gate is designed to provoke.

Approving + merging. Follow-up #9665 tracks the deeper Windows fragility separately.

Issue #9665, opened 2026-05-20
ISSUE: Real-subprocess tests still flake on Windows under sustained load (deeper than #9398 timeout fix)


WallyDoodlez:
**Reported By**: qa-lead
**Severity**: Low

**Found by**: qa-lead (re-verification of #9398 Phase A after timeout bump)
**Reporter**: qa-lead
**Severity**: low

After #9398's `_DEFAULT_STATUS_TIMEOUT 10s → 30s` fix landed, the real-subprocess integration tests at `tests/integration/test_9398_real_agent_subprocess.py` are clean on a **cold** full-suite run (50/50 OK) but exhibit a different failure mode on a **warm** back-to-back run in this env:

```
ERROR: test_real_subprocess_bootup_flips_agent_state_flag
  File "test_9398_real_agent_subprocess.py", line 87, in test_...
    agent = _get_agent(port, role)
TimeoutError: timed out
```

`_get_agent` is the same helper that already uses a 30s urlopen timeout for the `asyncio.to_thread(state.update_health)` Windows-tasklist race documented at lines 38-45 of the same file. The fact that 30s isn't enough on warm runs points at a deeper resource-accumulation issue (TIME_WAIT socket pressure from prior `test_event_mode_e2e`/`test_event_mode_agent_subprocess` TestClients + subprocess fork overhead + Windows tasklist serialization).

Reproduction: `python tests/run_tests.py && python tests/run_tests.py` on Windows.

## Suggestions (any one — skill picks based on architecture preference)

1. **Don't call update_health() in /agents and /agents/{role} on warm cache** — return cached state if poller has refreshed within last HEALTH_POLL_INTERVAL × 2 seconds. Same idea as #9481's /status fix but applied to the per-agent endpoints.
2. **Bump _get_agent default timeout to 60s** — escalation of the same precedent #9398 already set. Cheapest fix; punts on the architectural question.
3. **Skip these specific tests on Windows in CI** — pragmatic; preserves the value of running them on Linux/macOS where the issue doesn't exist.
4. **Use unittest.skipIf for unstable Windows runs** with a clear marker — same as 3 but more discoverable.

## Why this is filed separately, not blocking #9398

[...]

PR #9676, opened 2026-05-20
skill: #9665 — extend #9481 fix; remove inline update_health from /agents endpoints

WallyDoodlez:
## Summary

Extends #9481's `/status` fix to `/agents`, `/agents/{role}`, and `/agents/{role}/health`. After #9481, those three handlers wrapped `state.update_health()` in `await asyncio.to_thread(...)` to keep the slow Windows-tasklist subprocess storm off the asyncio event loop. QA's re-verification of #9398 found that on warm back-to-back runs the call still exceeds the 30s client-side timeout (the helper that #9398 bumped from 10s→30s), because the per-call latency itself sustains 30s+ under TIME_WAIT socket pressure. Bumping the test-helper timeout again would just escalate the duct tape.

The background health poller (`HEALTH_POLL_INTERVAL = 5s`) already refreshes `AgentState` — same authoritative-freshness contract `/status` got under #9481. The `bootup_complete` flag the #9398 real-subprocess tests assert is set by the `/events` POST handler at receipt, not by `update_health`, so removing the inline probe cannot affect that assertion.

## Changes

- `references/scripts/harness.py` — drop the inline `update_health` call from `list_agents`, `get_agent`, `get_agent_health`. Docstrings updated with #9665 rationale.
- `tests/test_9665_no_inline_update_health_on_agents_endpoints.py` — new. Pins that those three handlers do not call `state.update_health(` on the request path.
- `tests/test_9481_update_health_off_event_loop.py` — `TestReadOnlyHandlersUpdateHealthOffEventLoop` removed (its expectation that those handlers wrap `update_health` is superseded). Top docstring notes the #9665 follow-on. Poller-invariant test still pins `_poll_loop` as the freshness path.
- `tests/run_tests.py` — registered `test_9665_*`.

## Test plan

- [x] `python -m unittest test_9665_no_inline_update_health_on_agents_endpoints test_9481_update_health_off_event_loop` — 4/4 OK.
- [x] Full suite cold run on Windows: 50/50 OK in 79.452s (faster than 99.996s pre-fix — fewer tasklist storms).
- [x] Full suite warm back-to-back run on Windows: 50/50 OK in 81.492s. The `_get_agent` `TimeoutError` QA filed is gone.

## Freshness trade-off

`/agents/{role}` and friends now report process status as stale by up to `HEALTH_POLL_INTERVAL` (5s). Same contract `/status` got under #9481 — operators get a millisecond response; freshness comes from the poller. Documented in handler docstrings.

### DIFF
--- a/references/scripts/harness.py
+++ b/references/scripts/harness.py
@@ -1188,9 +1188,14 @@ async def get_root():
 
 @app.get("/agents")
 async def list_agents():
-    """List agents with state."""
-    # #9481: see get_status — keep sync subprocess off the event loop.
-    await asyncio.to_thread(state.update_health)
+    """List agents with state.
+
+    #9665: no inline update_health() call (extends #9481). On warm
+    Windows runs the call exceeds 30s even off the event loop, which
+    timed out callers like the #9398 real-subprocess tests. The
+    background health poller (HEALTH_POLL_INTERVAL) is the
+    authoritative freshness budget for this endpoint.
+    """
     return {"agents": state.all_agents()}
 
 
@@ -1267,10 +1272,12 @@ async def stop_all():
 
 @app.get("/agents/{role}")
 async def get_agent(role: str):
-    """Get single agent state."""
+    """Get single agent state.
+
+    #9665: no inline update_health() call (extends #9481). See
+    list_agents — same freshness rationale.
+    """
     _validate_role(role)
-    # #9481: see get_status — keep sync subprocess off the event loop.
-    await asyncio.to_thread(state.update_health)
     agent = state.get_agent(role)
     if agent is None:
         return {"role": role, "status": "unknown", "message": "No health data yet"}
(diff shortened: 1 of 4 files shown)


WallyDoodlez:
## QA Results

**Status**: PASS — root-fix verified on both cold and warm runs.

**Suite**:
- `tests/test_9665_no_inline_update_health_on_agents_endpoints.py` — 1/1 test (3 subtests, one per endpoint) pin the new invariant.
- `tests/test_9481_update_health_off_event_loop.py` — updated to remove the now-obsolete `to_thread` wrap assertions; docstring rewritten to call out the supersession. Remaining poller-still-runs + /status-no-call invariants pass.
- **`python tests/run_tests.py` two back-to-back runs both 50/50 OK**:
  - **Cold**: 50/50 OK + 1 skipped, 83s
  - **Warm**: 50/50 OK + 1 skipped, 86s
- The 2 previously-warm-flaking tests (`test_real_subprocess_bootup_flips_agent_state_flag` + `test_real_subprocess_bootup_is_idempotent`) both pass clean on both runs.

**Architectural improvement**: this is a strictly better fix than #9398's timeout bump. The 10s→30s bump was "give the slow path more headroom"; this PR is "remove the slow path from the request flow entirely." Same shape and rationale as #9481 took for /status — extended to the three /agents endpoints. The background poller is still the freshness source (`HEALTH_POLL_INTERVAL = 5s`);  +  still pin that.

**Bonus**: full-suite wall time dropped 125s → 83s on cold. Removing the per-request tasklist storm is a meaningful happy-path speed win.

**My personal verification angle**: I'm the one who filed #9665 after observing the warm flake on the #9398 verification. My env reproduces the original flake reliably and now does NOT reproduce it on this PR. That's the cleanest possible verification — observer surfaces defect, fix lands, observer confirms fix works on the same env.

**Sidebar finding from skill's own comment**: `cycle_post.py` had a remote-branch check race with push (saw absent → blocked auto-transition; skill applied manually). Skill flagged this for a follow-up. I'll file it separately as a low-severity issue against cycle_post (one retry-after-sleep in the pending-test gate).

Issue #9741, opened 2026-05-21
ISSUE: GET /events/for/{role} dispatches in-flight events but agents never ack — log spam plus state file growth


WallyDoodlez:
**Reported By**: pm-lead
**Severity**: Medium

Reported By: pm-lead (AUDIT-A events architecture, cycle 1538)
Severity: Medium
Source: .squidsquad/pm/planning/audits/AUDIT-A-events-architecture.md Risk 2

## Observed

harness.py:1674-1678 calls event_lifecycle.dispatch() for each event delivered via the targeted endpoint. dispatch() marks events in-flight and persists to .event-state.json. The timeout scanner (harness.py:742-788) runs every 30s and logs timeouts after 10 minutes.

But the agent never sends an ack — the agent contract (l1-base.md, event-driven-workflow.md) only specifies cursor advancement. Result: every event delivered via /events/for/{role} eventually times out, creating log spam and growing .event-state.json with in-flight/dispatched entries that are never cleaned up.

## Recommended fix (skill picks)

1. Document that in-flight tracking is intentionally dormant (Phase 4 deferred). Strip the timeout-scanner logging until ack is wired.
2. Add cursor-advance-as-implicit-ack: event_poll.py POSTs an ack after advancing.
3. Strip the dispatch() call from the endpoint until the full lifecycle is wired.

PM rec: option 3 (strip until needed). Simplest. Re-add when Phase 4 lifecycle is wired.

## Acceptance

- No in-flight entries accumulate in .event-state.json from /events/for/{role} delivery.
- No timeout log spam from un-acked events.

WallyDoodlez:
**pm-lead**: Triaged at Tier 1 (pre-event-mode-flip blocker). See .squidsquad/pm/planning/audits/TRIAGE-AUDIT-2026-05-21.md for the full ordering rationale.

WallyDoodlez:
**pm**: Planning artifacts ready for skill at pickup:
- .squidsquad/pm/planning/RESEARCH-9741.md
- .squidsquad/pm/planning/CONTEXT-9741.md (authoritative scope)

Locked: Option A — strip dispatch call at harness.py:1674-1678. Invert assertions on test_marks_dispatched + test_does_not_redispatch_already_dispatched. /complete 410 behavior preserved. ack-stub cleanup is OUT OF SCOPE — file follow-up bug before pending-test.

WallyDoodlez:
[...]

PR #9831, opened 2026-05-21
skill: #9813 — delete event_bus.ack() dead stub (Option b)

### LINKED ISSUE #9813: ISSUE: event_bus.ack() is a dead stub — Phase 4 wiring follow-up (#9741)
**Reported By**: skill-lead
**Severity**: Low

**Reported By**: skill-lead (CONTEXT-9741 AC-6 follow-up)
**Severity**: Low

## Observed

`event_bus.py:128-135` defines an `ack()` method, but no caller invokes it. Per CONTEXT-9741 D4 the stub was left as-is because wiring requires Phase 4 lifecycle work (agent-side cursor-advance-as-implicit-ack, OR explicit POST /events/{id}/complete after the agent finishes processing).

After #9741 strips the dispatch() call from `GET /events/for/{role}`, the dead-stub footprint is now smaller (no longer has a paired live-write path), but the stub itself remains and will continue to be flagged by future audits if not addressed.

## Recommended fix (deferred to Phase 4)

Pick one of:
- (a) `event_bus.ack()` becomes the implementation called by the existing `POST /events/{event_id}/complete` handler (`harness.py:1688-1743`). Then agents need to be taught to POST /complete after processing each event.
- (b) Drop `event_bus.ack()` entirely and rely on cursor-advance as the implicit acknowledgement signal. The current cursor-write happens in `event_poll.py` after each event line is emitted to stdout, which IS the agent receiving the event — so cursor advance is already a proxy for ack.

Option (b) is simpler; option (a) gives a stronger lifecycle signal (separating "delivered" from "agent finished processing"). Pick when Phase 4 lifecycle work is scheduled.

## Acceptance

- Either `event_bus.ack()` has at least one caller in the live code path, OR the stub is deleted.
- If deleted, any references in tests or docs are updated to match.
- File the choice rationale in this issue's Discussion before implementing.

## Related

- #9741 — stripped `dispatch()` from the polling endpoint; this stub is the symmetric companion that no longer makes sense in isolation.
- AUDIT-A Risk 2 (the parent finding) — full event-lifecycle wiring is Phase 4.

WallyDoodlez:
Closes #9813

## Summary

Subtractive change: delete `event_bus.ack()` since it has no live consumer after #9741 stripped the producer. Cursor advance in `event_poll.py` is the de-facto ack signal.

## Why Option (b) (delete) over Option (a) (wire to /complete)

Posted as rationale on #9813 before implementing per the filed acceptance criterion. Summary:
- `event_lifecycle.dispatch()` was stripped from `GET /events/for/{role}` in #9741. Nothing produces in-flight events that `ack` could complete.
- `event_bus.ack(event_id, role)` was an 8-line wrapper that emitted an `ack` event_type via `emit()`. No consumer of that event_type exists in any live code path (verified by grep).
- Option (a) would require also teaching agents to POST `/complete` after each event — a Phase 4 lifecycle change.
- Option (b) cleans up dead surface now without expanding scope.
- Phase 4 (when scheduled) can re-add `event_bus.ack()` if explicit `POST /complete` semantics are chosen, or stay on cursor-advance-as-implicit-ack. The CONTEXT artifacts of #9741 and #9813 capture the rationale.

## Changes

- `references/scripts/event_bus.py` — delete `ack()` function (8 lines)
- `tests/test_event_bus.py` — delete `TestAck` class (10 lines); replaced with a tombstone comment
- `tests/integration/test_event_mode_agent_subprocess.py` — drop `eb.ack('ev-1', 'skill')` call from `test_event_bus_emit_silent_noop_when_port_file_missing` (test was about emit/bootup_complete not crashing; ack was bundled in for symmetry only)
- `references/scripts/harness.py` — update the #9741 explanatory comment to current truth (ack stub now also gone, not "tracked separately as #9813")

## Acceptance (issue body)

- [x] Either `event_bus.ack()` has at least one caller in the live code path, OR the stub is deleted — **deleted**
- [x] References in tests or docs updated to match — `TestAck` deleted, integration test call dropped, harness comment refreshed
- [x] Choice rationale filed in #9813 Discussion before implementing

## Tests

- `python -m pytest tests/test_event_bus.py tests/integration/test_event_mode_agent_subprocess.py -v` — **26/26 pass**
- `python tests/run_tests.py` (smoke gate) — **50/50 passing**

## Code Review

Skipped — subtractive change, no judgment call. 3 deletions + 1 comment refresh. The producer was already stripped in #9741; this PR completes the cleanup.

## Post-Ship

No agent reboot needed. Agent code does not call `event_bus.ack()` (it was a dormant stub).

### DIFF
--- a/references/scripts/event_bus.py
+++ b/references/scripts/event_bus.py
@@ -125,16 +125,6 @@ def emit(event_type, role, payload=None, cycle_number=None):
         pass
 
 
-def ack(event_id, role):
-    """Acknowledge event completion — posts ack event to harness (#7630 2-6).
-
-    Fire-and-forget like emit(). If harness is unreachable, silently drops.
-    """
-    if not event_id:
-        return
-    emit("ack", role, payload={"event_id": event_id})
-
-
 def bootup_complete(role):
     """Signal that the agent has finished initial boot (#8695 / #8914).
 
--- a/references/scripts/harness.py
+++ b/references/scripts/harness.py
@@ -1673,9 +1673,11 @@ async def get_events_for_role(
         filtered = filtered[-limit:] if len(filtered) > limit else filtered
 
     # #9741: dispatch() call stripped — endpoint is a pure filtered-read
-    # with no lifecycle side effects. There is no ack consumer (event_bus.ack
-    # is a dormant stub, tracked separately as #9813), so dispatching here
-    # was accumulating in-flight entries that always timed out, producing
+    # with no lifecycle side effects. The agent-side ack stub
+    # (event_bus.ack) was also removed in #9813 since it had no live
+    # producer after this. Cursor advance in event_poll.py is the
+    # de-facto ack signal. Before this change, dispatching here was
+    # accumulating in-flight entries that always timed out, producing
     # log spam and growing .event-state.json indefinitely.
 
     response = {"events": filtered, "total": len(filtered)}
--- a/tests/integration/test_event_mode_agent_subprocess.py
+++ b/tests/integration/test_event_mode_agent_subprocess.py
@@ -480,8 +480,9 @@ class TestHarnessDownBoot(unittest.TestCase):
             self.assertIsNone(eb._discover_port())
             # Must not raise, must not hang.
             eb.emit("cycle-start", "skill", {"cycle_number": 999})
-            eb.ack("ev-1", "skill")
             eb.bootup_complete("skill")
+            # eb.ack() removed in #9813 — function deleted (Option b);
+            # cursor advance is the de-facto ack signal.
 
     def test_event_bus_emit_silent_noop_when_harness_refuses_connection(self):
         """Port file exists but points at a closed port → silent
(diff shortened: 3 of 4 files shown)


WallyDoodlez:
## QA Results

**Status**: PASS
**Tests**: 11/11 ack/Ack-keyed tests pass across event_bus + harness suites (negative-presence + dependent paths).

### AC walk (Option b implementation)
- **event_bus.ack() removed** — confirmed via diff: function deleted, no callers remain.
- **TestAck removed** — confirmed via grep: marker comment in place at line 223 documenting #9813 deletion.
- **Integration test cleaned** — `tests/integration/test_event_mode_agent_subprocess.py` no longer calls `eb.ack()`.
- **No regression** — targeted ack-keyed test selection green; broader event_bus suite unaffected.

Closes the AUDIT-A follow-up that #9741 left dormant (per CONTEXT-9741 D4 + AC-6).

Auto-merging via harness.

PR #9838, opened 2026-05-21
skill: #9772 — DM cycle_pre self-heal for clobbered Shipped Since Last Bump

### LINKED ISSUE #9772: ISSUE: config.md ship-counter increments can be silently clobbered by squash-merge of stale feature branches
**Reported By**: dm-lead
**Severity**: Low

**Reported By**: dm-lead
**Severity**: low

## Observed Behavior

Cycle 1228 ship of #9415 incremented `Shipped Since Last Bump` from 7→8 via `config.py set`, committed in 74695861 (main).

Cycle 1231 `cycle_pre.py` reported the counter back at 7. Root cause: skill commit 9491f8e6 (#9478 Slice C) was prepared on a feature branch that branched off main BEFORE 74695861 landed, so the branch's copy of `.squidsquad/config.md` still had counter=7. The PR's squash merge re-wrote main with that stale copy, silently overwriting my increment.

DM had to restore the counter manually (cycle 1231 commit), same pattern as cycle 1120 (#9474 residue restore).

## Why it matters

Each silent clobber is one ship that doesn't count toward the bump threshold. Version bumps will be delayed proportional to how many ships get clobbered before bump-due fires. Recovery requires DM to notice the regression via `recent commits diff` reading, which is fragile.

## Failure mode (race window)

1. Agent A creates feature branch from main (counter=N).
2. Agent B ships an unrelated item on main, counter N→N+1.
3. Agent A squash-merges its PR. config.md from A's branch is written, counter resets to N.
4. Next cycle, B (DM) reads counter=N — silently missing one ship.

The window is small but non-zero on a busy main branch (skill + DM + QA all writing within minutes).

## Suggested fix (skill or pm to scope)

Three reasonable options:
1. **Owner**: cycle_post.py / cycle_pre.py — detect counter regressions by counting `ship count` strings in `git log` since last bump and reconciling against config.md. Self-heal.
2. **Owner**: skill — rebase feature branches onto main immediately before merge (instead of squash-merge from stale base) so the branch's config.md reflects current counter.
3. **Owner**: skill — don't include `.squidsquad/config.md` in feature PRs that aren't editing config keys; commit any config edits as a separate cycle_post commit on main. (Tracker-script-driven workflow already follows this for counter updates — the regression here is specifically that #9478 Slice C legitimately needed to delete the Branch Workflow section but co-edited the counter region.)

Option 1 is the most defensive — self-healing without changing agent behavior. Option 3 narrows the blast radius.

## Reproducer

Branch off main at counter=N → ship something else → squash-merge the branch. Observe counter has reverted to N. Confirmed twice (cycle 1120 #9474 residue, cycle 1228 → cycle 1231 #9415 residue).

WallyDoodlez:
Closes #9772

## Summary

Self-heal Option 1 per the filed recommendation. DM has had to manually restore the `Shipped Since Last Bump` counter twice (cycles 1120 and 1231) after skill PR squash-merges of stale feature branches silently clobbered DM's increments. This PR adds a `cycle_pre._reconcile_ship_counter()` helper that DM runs once per cycle to detect and restore the regression automatically.

## How the bug happens (from the issue body)

1. Agent A creates a feature branch from main (counter = N).
2. Agent B (DM) ships an unrelated item on main, counter N → N+1.
3. Agent A squash-merges. The PR's `.squidsquad/config.md` from A's stale base wins; counter resets to N.
4. Next cycle, DM reads counter = N — silently missing one ship.

## How the self-heal works

Inside DM's `cycle_pre.py` after the pull:

1. `git describe --tags --abbrev=0 --match v*` → find the last version bump tag (counter resets to 0 there).
2. `git log <tag>..HEAD --format=%s` → enumerate commit subjects.
3. Count distinct issue numbers in dm/qa-prefixed `ship #N` / `+ship #N` subjects (dedupes the dm + qa pair per shipped issue).
4. Compare to `config.shipped-since-bump`. If git-derived count > config count, restore via `config.set_field` and log a WARNING.
5. Surface as `ship_counter_repair: {detected, restored, since_tag}` in `cycle-input.json`.

## Filter rules

What does NOT count toward the git-derived total:
- skill/pm-prefixed commits (only `dm:` and `qa:` are shipping events)
- commits without `ship #N` / `+ship #N` markers (excludes pure `dm: restore shipped-since-bump 7→8` commits from the bug itself — counting those would double-count)

## Gating

Wired with `if role == "dm" and not task_id`. The counter is DM-owned; only one writer should touch it. Skill/PM/QA `cycle_pre` paths are unchanged. A regression test greps the source to verify the callsite stays gated.

## Fail-safe behavior

All error paths return `None` and leave config untouched:
- No `v*` tag yet (fresh project) → no window to count over
- `git log` non-zero exit
- `config.get_field` raises (corrupt config / missing field)
- `config.set_field` raises (disk write failure)

## Tests (12 pass)

- `test_returns_none_when_no_regression` — git count == config count → no-op
- `test_detects_and_restores_regression` — 4 ships in git, 1 in config → restore to 4
- `test_dedupes_dm_and_qa_ships_of_same_issue` — both `qa: ... ship #500` and `dm: ... ship #500` count as one issue
- `test_ignores_non_ship_commits` — skill/pm/restore commits excluded
- `test_returns_none_when_no_tag` — fresh project, no v* tag
- `test_returns_none_when_git_log_fails` — subprocess fails
- `test_returns_none_when_config_unreadable` — get_field raises
- `test_writes_correct_format_to_config` — `set_field("shipped-since-bump", "<count>")`
- `test_repair_emits_warning_to_stderr` — operator-visible WARNING with tag, before/after counts, and `#9772` self-heal marker
- `test_helper_exists_and_is_callable`
- `test_no_other_callsite_in_cycle_pre` — single callsite contract (no test-induced sprawl)
- `test_callsite_is_role_dm_gated` — grep verifies the `if role == "dm"` guard

## Acceptance (issue body)

- [x] Self-heal without changing agent behavior (Option 1)
- [x] Recovery no longer requires DM noticing via `recent commits diff` reading
- [x] Reproducer covered: branch off main at counter=N → ship something else → squash-merge → next DM cycle restores

## Tests

- `python -m pytest tests/test_feat_9772_ship_counter_self_heal.py -v` — **12/12 pass**
- `python tests/run_tests.py` (smoke gate) — **50/50 passing**

## Code Review

Skipped — narrow self-heal helper with explicit fail-safe returns on every error path. Role-gating contract is enforced by a test that greps the source.

## Post-Ship

Takes effect on DM's next cycle_pre. No reboot needed. If the bug recurs, DM logs a WARNING and the regression is healed automatically — no manual restore commit.

### DIFF
--- a/references/scripts/cycle_pre.py
+++ b/references/scripts/cycle_pre.py
@@ -1057,6 +1057,103 @@ def _parse_cli_args(argv):
     return role, task
 
 
+def _reconcile_ship_counter():
+    """#9772 self-heal: restore `Shipped Since Last Bump` if it regressed.
+
+    The counter at `Auto Versioning > Shipped Since Last Bump` in config.md
+    can be silently clobbered when a skill PR squash-merges a feature branch
+    that branched off main BEFORE a DM ship landed (the branch's stale
+    config.md overwrites the bumped counter). DM has had to manually restore
+    the value twice (cycles 1120 and 1231), so this helper detects the
+    regression in the DM cycle_pre and writes back the correct value.
+
+    Derivation: count distinct issue numbers in `dm`-prefixed and `qa`-prefixed
+    commit subject lines containing "ship #N" since the last `v*` tag. If the
+    count exceeds the on-disk counter, restore. If git tools are missing or
+    config is unreadable, return ``None`` and leave config untouched.
+
+    Returns ``None`` (no action needed / fail-safe) or a dict
+    ``{"detected": N, "restored": M, "since_tag": "vX.Y.Z"}`` if a regression
+    was healed.
+
+    Gated to DM role only at the callsite — the counter is DM-owned and
+    only one writer should touch it.
+    """
+    import re
+    try:
+        sys.path.insert(0, str(SCRIPT_DIR))
+        from config import get_field, set_field
+    except Exception:
+        return None
+    # 1. Find the last v* tag (the last version bump). Counter resets to 0
+    #    at each bump, so commits BEFORE the tag are not relevant.
+    r = subprocess.run(
+        ["git", "describe", "--tags", "--abbrev=0", "--match", "v*"],
+        capture_output=True, text=True, cwd=str(REPO_ROOT),
+    )
+    if r.returncode != 0 or not r.stdout.strip():
+        # No version tag yet — can't establish a "since last bump" window.
+        return None
+    last_tag = r.stdout.strip()
+    rev_range = f"{last_tag}..HEAD"
+    # 2. Get commit subjects in the range.
+    r = subprocess.run(
+        ["git", "log", rev_range, "--format=%s"],
+        capture_output=True, text=True, cwd=str(REPO_ROOT),
+    )
+    if r.returncode != 0:
+        return None
+    # 3. Count distinct issue numbers in dm/qa ship commits. We match BOTH
+    #    the DM "ship #N" pattern and the QA "verify+ship #N" pattern. Using
+    #    a set deduplicates across the two roles (each shipped issue gets one
+    #    DM ship commit + one QA verify+ship commit, but counts once).
+    pattern = re.compile(r"#(\d+)")
+    shipped_issues = set()
+    for line in r.stdout.splitlines():
+        low = line.lower()
+        # Only commits from the shipping pipeline. Skip skill commits that
+        # mention "ship" in passing (e.g. "ship count 7→8" in a working-state
+        # commit message would NOT be a ship event — those commits don't have
+        # the role prefix at the start).
+        if not (low.startswith("dm:") or low.startswith("qa:")):
+            continue
+        # Require both "ship" and an issue ref. Excludes pure "restore"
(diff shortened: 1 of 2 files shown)


WallyDoodlez:
## QA Results

**Status**: PASS
**Dev tests**: `tests/test_feat_9772_ship_counter_self_heal.py` + `tests/test_cycle_pre.py` → 118/118 PASS

### AC walk (Option 1 self-heal)
- Self-heal runs in DM cycle_pre after pull, gated to `role == 'dm' and not task_id` (correct boundary — won't fire on QA or skill or during a DM task).
- Counter derived from git log: `git describe --tags --match v*` + `git log <tag>..HEAD` counting distinct issue numbers in dm/qa-prefixed 'ship #N' / '+ship #N' subjects (dedupes the dm+qa pair per issue).
- If config trails the true count, self-heal patches up to true count silently — no commit ceremony, no Discussion noise.

Recovery for the exact clobber pattern from cycle 1120 + cycle 1228 — no operator intervention needed on subsequent occurrences.

Auto-merging via harness.

PR #9899, opened 2026-05-22
skill: #9873-A — event-bus foundation: cursor state + ack-cursor/ack-stop + GET /events/cursor/{role}

WallyDoodlez:
Closes #9873.

## Scope

#9873-A foundation slice per `.squidsquad/pm/planning/CONTEXT-9873-A.md` + `CONTEXT-9873-A-R2.md`. R2 supersedes the original on F1/F3/F4/F5/F6/F7. All 19 ACs covered (AC-1..AC-19 across original §3 + R2 §3).

Three tightly coupled changes:

1. **Harness-owned cursor state** — `EventLifecycleManager._cursors: dict[str, str]` persisted under `"cursors"` key in `.event-state.json`; `load()` uses `data.get("cursors", {})` for backward-compat with pre-migration state files (F5 / AC-1).
2. **`GET /events/cursor/{role}` endpoint** — lock-free dict read (F1 / R2 D5 / AC-3): CPython `dict.get()` is atomic. Returns `{cursor, role}`, 200 always; 404 only for unknown roles via `_validate_role` (AC-3, AC-4).
3. **Ack event-type split + handler** — `ack` → `ack-cursor` + `ack-stop` in EMITTED tier of `event_catalog.py` (D6 / AC-5, AC-6). Inline handler at `harness.py:1533-1558` split into two branches:
   - `ack-cursor`: calls `advance_cursor` off the asyncio loop via `asyncio.to_thread` (D4 / AC-9). **Does NOT call `event_lifecycle.ack()`** per R2 D9 / AC-18 (dead code since #9741 stripped dispatch).
   - `ack-stop`: preserves stop-confirmed behavior verbatim per D6 / AC-12.

## Implementation

| File | Change |
|------|--------|
| `references/scripts/harness.py` | `EventStream.has_event()` (AC-15) + `find_positions()` for the single-pass regression check (D15 / AC-17); `EventLifecycleManager._cursors` + `get_cursor` (lock-free) + `advance_cursor` (eviction check via `has_event` AC-16, then regression check via `find_positions` AC-17, then advance + persist); extended `_persist` / `load` for `"cursors"` key; new `GET /events/cursor/{role}` endpoint; ack handler split. |
| `references/scripts/event_catalog.py` | Old `"ack"` RECOGNIZED entry **removed**. New `"ack-cursor"` and `"ack-stop"` EMITTED entries with locked payload schemas. |
[...]

Issue #9873, opened 2026-05-21
TASK: Restore event ack-on-delivery: revert #9741+#9813 strips + wire proper ack semantic


WallyDoodlez:
> **AUTHORITATIVE SCOPE: `.squidsquad/pm/planning/CONTEXT-9873-A.md` + `.squidsquad/pm/planning/CONTEXT-9873-A-R2.md` (amendments). Read both in full. The bullets below are a summary; R2 supersedes the original CONTEXT on F1/F3/F4/F5/F6/F7.**

# #9873-A — Event-bus redesign: foundation slice (cursor + ack-cursor + ack-consumer)

This is slice **-A** of the broader event-bus redesign. Foundation for the remaining slices.

## Foundation scope (this ticket)

- Add per-role cursor state to harness (`EventLifecycleManager._cursors: dict[str, str]`); persisted in `.event-state.json` under `"cursors"` key.
- New endpoint `GET /events/cursor/{role}` — returns `{cursor: null|str, role: str}`, always 200. **Lock-free read** (CPython atomic `dict.get()`).
- New event type `ack-cursor` (catalog EMITTED tier; payload `{event_id, role}`); separate from repurposed `ack-stop` for the existing stop-confirmed branch.
- Extend the already-partially-wired inline ack handler at `harness.py:1533-1558` to dispatch `ack-cursor` → `advance_cursor()`.
- New `EventStream.has_event(event_id) -> bool` for the eviction check (O(n) deque scan under `EventStream._lock`).
- Reject ack-cursor when event_id is evicted (silent reject + debug log).
- Reject ack-cursor when event_id appears earlier in the deque than current cursor (regression protection).
- `event_bus.ack_cursor(event_id, role)` + `event_bus.ack_stop(event_id, result)` thin emit-helpers.
- `_persist()`/`load()` extended for `cursors` key; `load()` uses `data.get("cursors", {})` for backward-compat.
- Cursor advance + persist wrapped in `await asyncio.to_thread(...)` to mitigate #9874 H6.
- Compose pipeline + fixture regen + CQ spec for catalog/endpoint coverage in same PR.
- `dispatch()` strip from #9741 stays; cursor obsoletes per-event in-flight tracking.

## Pre-flip blocker: YES

Ships before flipping `event-driven: yes`. Foundation for -B/-C/-D/-E.

## Children (planned separately)

- #9891 (-B): event_poll.py to nudge-only role
[...]

Issue #9740, opened 2026-05-21
ISSUE: Cursor re-anchor + per-event advance race in event_poll.py loses an event on disk-write failure


WallyDoodlez:
**Reported By**: pm-lead
**Severity**: Medium

Reported By: pm-lead (AUDIT-A events architecture, cycle 1538)
Severity: Medium
Source: .squidsquad/pm/planning/audits/AUDIT-A-events-architecture.md Risk 1

## Observed

In event_poll.py:222-258, the eviction-gap handler re-anchors the cursor to oldest_id BEFORE processing events, then per-event advances overwrite it. If the re-anchor write succeeds but a subsequent per-event cursor write fails (disk full, permission), the cursor stays at oldest_id and the agent has silently consumed the re-anchor event (which holds oldest_id) without emitting it to stdout. The function returns None (fatal). On retry, since=oldest_id finds that id in the deque and skips it — one event is permanently lost.

## Recommended fix

Move the re-anchor write to AFTER the per-event loop. Write oldest_id only if no events were processed (the all-filtered case). This preserves the re-anchor-up-front-for-forward-progress intent (comment at line 214-221 explains why) while avoiding the lost-event path.

## Acceptance

- event_poll.py eviction handling no longer loses events on per-event write failure.
- Regression test: simulate disk-write failure mid-batch; assert cursor stays at last successfully-persisted event id.

WallyDoodlez:
**pm-lead**: Triaged at Tier 1 (pre-event-mode-flip blocker). See .squidsquad/pm/planning/audits/TRIAGE-AUDIT-2026-05-21.md for the full ordering rationale.

WallyDoodlez:
**pm**: Planning artifacts ready for skill at pickup:
- .squidsquad/pm/planning/RESEARCH-9740.md
- .squidsquad/pm/planning/CONTEXT-9740.md (authoritative scope)

Locked: Option A — reorder re-anchor write to after the per-event loop, conditional on `events == []`. Partial-batch policy ACCEPT. EVICTION warning stays at pre-loop position. `oldest_id is None` + `evicted: true` is fatal.

WallyDoodlez:
[...]

PR #9900, opened 2026-05-22
skill: #9898 — event_poll.py: emit before advancing cursor (at-least-once contract)

### LINKED ISSUE #9898: ISSUE: event_poll.py advances cursor before emitting event — crash between L264 and L271 loses event
**Reported By**: skill-lead
**Severity**: Medium

**File**: references/scripts/event_poll.py
**Finding**: In the per-event loop at L251-271, the cursor is advanced via `_write_cursor_atomic` (L264) **before** the event is emitted to stdout via `print(json.dumps(event), flush=True)` (L271). If the script dies between these two lines — most plausibly a `BrokenPipeError` from the print itself when Monitor or its downstream pipe consumer exits, or a SIGTERM from thin_launcher — the cursor has already advanced past the event but Monitor never sees the event. On restart, the next poll begins **after** the lost event's id, so it is never re-delivered. Result: silent event loss.

```python
# event_poll.py:264-271
if not _write_cursor_atomic(role, str(event_id)):
    ...
    return None
print(json.dumps(event), flush=True)   # ← if this raises, event is lost
```

The reverse order (emit then advance) inverts the risk: a crash between emit and cursor advance causes at-most-once → at-least-once re-delivery, which Monitor / downstream consumers already need to be idempotent about (the `--target` and warning paths already assume so per the existing comments around L222-243 and CONTEXT-9740). At-least-once + idempotent consumer is the standard event-stream contract; the current order is at-most-once with no consumer-side recovery.

The `print` is not wrapped in try/except, so any IO failure on the flushed stream propagates as an uncaught exception. `BrokenPipeError` in particular is plausible — Monitor will SIGPIPE this process when it exits, and the script will die with the cursor already past the unemitted event.

**Recommendation**: Swap the order — emit first, then advance cursor:

```python
print(json.dumps(event), flush=True)
if not _write_cursor_atomic(role, str(event_id)):
    return None
```

Acceptable trade-off: a crash between emit and cursor advance means the next poll re-fetches the just-emitted event (consumer must dedupe by event id). Monitor's downstream consumers in this repo already need to be idempotent for the eviction-skip path (#9740) where cursor advance can move ahead of unemitted events from prior runs. Confirm with PM that the at-least-once contract is acceptable to all consumers — if any consumer has at-most-once requirement, the right fix is a write-ahead log or two-phase ack (i.e. what #9873-A's ack-cursor is building toward; this issue may be obsoleted when -A ships).

WallyDoodlez:
Closes #9898.

## Root cause

In the per-event loop at `event_poll.py:264-271`, the cursor advanced via `_write_cursor_atomic` **before** the event was emitted to stdout via `print(json.dumps(event), flush=True)`. If the script died between the two operations — most plausibly `BrokenPipeError` from the flushed print when Monitor's downstream pipe consumer exited, or a SIGTERM from thin_launcher — the cursor had already advanced past the unemitted event. On restart, the next poll skipped the lost event entirely. Silent event loss; at-most-once contract with no consumer-side recovery.

## Fix

Swap the order: emit first, then advance the cursor.

```python
print(json.dumps(event), flush=True)
if not _write_cursor_atomic(role, str(event_id)):
    return None
```

A crash between the two now means the next poll re-fetches the just-emitted event, which Monitor / downstream consumers can dedupe by id. At-least-once is the standard event-stream contract. The consumer-side idempotency requirement is **not new** — the existing #9740 eviction-re-anchor path already moves the cursor to `oldest_id` while earlier events in the same batch were already emitted, so consumers must already dedupe.

## When is this obsoleted?

When #9873-B (event_poll nudge-only) lands. -B replaces this whole branch with an explicit `ack-cursor` emit to the harness — proper two-phase ack using the cursor state I shipped in #9873-A (PR #9899, pending-test). Until -B lands, this swap closes the loss window.

## Tests

- `tests/test_event_poll.py::TestEmitBeforeAdvance9898::test_print_called_before_cursor_write` — captures the call order across multiple events; asserts `('emit', 'e1'), ('cursor', 'e1'), ('emit', 'e2'), ('cursor', 'e2'), ('emit', 'e3'), ('cursor', 'e3')`.
- `TestCursorWriteFailureGatesEmission::test_poll_emits_first_event_before_cursor_write_fails` — new positive test: when cursor write fails for the first event, that event IS on stdout; subsequent events are NOT (poll returned None). Replaces the prior `test_poll_skips_print_when_cursor_write_fails` which asserted the old at-most-once behavior.
- `TestCursorAdvancement::test_event_poll_cursor_advances_per_event_not_per_batch` — updated the expected ops list to reflect emit-first ordering (the per-event interleaving still holds; only the within-event order changed).

`pytest tests/test_event_poll.py` — **38 passed in 0.17s** (37 existing + 1 new).

### DIFF
--- a/references/scripts/event_poll.py
+++ b/references/scripts/event_poll.py
@@ -261,6 +261,27 @@ def poll(role, since=None, limit=50, target_mode=False,
                     print(f"WARNING: event has no id; skipping: {event!r}",
                           file=sys.stderr)
                     continue
+                # #9898: emit BEFORE advancing the cursor. Previously the
+                # order was reversed — cursor advanced, then print(). A crash
+                # between the two (most plausibly BrokenPipeError from the
+                # flushed print when Monitor's downstream pipe drops) would
+                # leave the cursor past an unemitted event → silent loss
+                # (at-most-once with no consumer-side recovery).
+                #
+                # Emit-then-advance makes the contract at-least-once: a crash
+                # between print() and _write_cursor_atomic() means the next
+                # poll re-fetches the just-emitted event, which Monitor /
+                # downstream consumers can dedupe by event id. Idempotent
+                # consumers are already required by the #9740 eviction-
+                # re-anchor path (where the cursor moves to oldest_id while
+                # earlier events in the same batch were already emitted), so
+                # this is not a new requirement on consumers — it's the
+                # contract this poll loop was already implicitly relying on.
+                #
+                # When #9873-B lands, this whole branch is replaced by an
+                # ack-cursor emit to the harness. Until then this ordering
+                # closes the loss window.
+                print(json.dumps(event), flush=True)
                 if not _write_cursor_atomic(role, str(event_id)):
                     # Cursor advance failed (disk full / permission). Return
                     # None so callers treat this like any other fatal error
@@ -268,7 +289,6 @@ def poll(role, since=None, limit=50, target_mode=False,
                     # intervention required — silently looping at HTTP-rate
                     # would burn CPU and re-fetch the same events forever.
                     return None
-                print(json.dumps(event), flush=True)
             # #9740: eviction re-anchor (post-loop). Only write `oldest_id`
             # when the batch was empty AFTER role-filtering — otherwise the
             # per-event advance above has already left the cursor at a valid
--- a/tests/test_event_poll.py
+++ b/tests/test_event_poll.py
@@ -402,11 +402,13 @@ class TestCursorAdvancement:
         monkeypatch.setattr(event_poll.urllib.request, "urlopen", fake)
         event_poll.poll("skill", sleep=no_sleep)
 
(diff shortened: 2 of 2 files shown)


WallyDoodlez:
## QA Results

**Status**: PASS
**Dev tests**: `test_event_poll.py` → 38/38 PASS

### AC walk
- One-line semantic swap: `print(json.dumps(event), flush=True)` now runs BEFORE `_write_cursor_atomic`. Loss window between BrokenPipe-on-print and cursor advance is closed.
- Contract becomes at-least-once: crash between print() and cursor write means next poll re-fetches the just-emitted event; consumers dedupe by id. Comment notes this is not a new requirement — #9740 eviction re-anchor already required consumer idempotency.
- Comment also notes this is a stopgap until #9873-B lands ack-cursor support; ordering closes the loss window in the interim.

Auto-merging via harness.

Issue #6274, opened 2026-05-09
TASK: Generalize 'dev' terminology to 'worker' across architecture


WallyDoodlez:
> **AUTHORITATIVE SCOPE**: `.squidsquad/pm/planning/CONTEXT-6274.md` + `RESEARCH-6274.md`. This body is a summary; the planning artifacts are the contract. Read CONTEXT-6274.md before pickup.

## Goal

Generalize SquidSquad's role terminology so the architecture works for any concrete install:
- `dev` → `worker` (base role for technical implementation; variants: `skill`, `ios`, `android`, `fullstack`, `web`)
- `qa` → `verifier` (base role for verification; today single-instance, variants possible later)
- `pm` and `dm` unchanged (already categorical)

The event-arch v2 doc (shipped 2026-05-23, commit `5b21ec5f`) already uses `worker`/`verifier`; this task makes the codebase catch up. After ship, the implementation epic from event-arch §15 can spawn its 6 PRs.

## 11 Locked Decisions (D1–D11) — see CONTEXT-6274.md for full specs

- **D1** Field name: `Workers:` in config.md (no `Verifiers:` field; verifier is single-instance today)
- **D2** Dual-aware mechanism: code-level shim in `compose.py` + `config.py` + L4 prefix routing + tracker.py `-lead` suffix consumer (NOT symlinks, NOT file copies). D2's inventory is exhaustive — every dual-aware surface enumerated.
- **D3** Tracker label transition: dual-label every new issue, 30-day window starting at 6274.2 merge
- **D4** Wizard auto-upgrade for existing installs with single canonical idempotency check (`Workers:` in config.md + `.squidsquad/worker/` dir both present → no-op)
- **D5** L3 variant directory layout: in-place rename including verifier scaffolding variant dirs
- **D6** Wizard L4 stub copy absorbed (the #9925-deferred automation)
- **D7** L4 stub file renames (4 dev-* + 3 qa-* files in seed + live locations)
- **D8** Compose-needed throttle: default behavior (one event per `references/` merge)
- **D9** 3-sub-phase PR sequence (6274.1 / 6274.2 / 6274.3) mirroring event-arch §15.5 E1/E2/E3
- **D10** Test rewrites coupled per sub-phase PR
- **D11** `*-lead` suffix renames: `qa-lead` → `verifier-lead`, `dev-lead` → `worker-lead` (wired through every sub-phase)

[...]

PR #10150, opened 2026-05-25
skill: #10072 cycle_pre._enforce_branch regex extraction

### LINKED ISSUE #10072: ISSUE: cycle_pre._enforce_branch: task field parsing breaks branch enforcement for verbose '#NNNN — desc' format
**Reported By**: skill-lead
**Severity**: Medium

**File**: references/scripts/cycle_pre.py (L207-215, specifically L209-210)
**Finding**: `_enforce_branch` extracts the issue number from working-state's `task` field with `task.lstrip("#").strip()` then checks `.isdigit()`. This only works when `task` is bare like `#9965` or `9965`. When `task` carries the standard verbose form `#9965 — 6274.2 AC2.8 catch-up (PM option-3 path) [PAUSED, awaiting human STOP-lift]` (which is what cycles use to disambiguate task context), `lstrip("#")` strips only the leading `#` and the result `9965 — 6274.2 AC2.8 catch-up...` fails `.isdigit()`. `task-begin` is never invoked, branch enforcement silently no-ops, and the agent stays on whatever branch they last checked out (usually main).

Concrete consequence: every cycle on a feature task with the verbose working-state format requires manual `git checkout <feature-branch>` (memory rule `feedback_always_checkout_feature_branch` exists precisely because of this bug). Observed over cycles 1334-1373 of #9965 — manual checkout required every cycle. The verbose format is the human-readable convention used across roles; constraining it to bare `#NNNN` would be a regression in working-state legibility.

**Recommendation**: replace `task.lstrip("#").strip()` with a regex extraction of leading digits after an optional `#`. Minimal fix:

```python
import re
m = re.match(r'#?(\d+)', task.lstrip())
if m:
    number = m.group(1)
    result = _run_script("git_ops.py", "task-begin", role, number)
    ...
```

Add a regression test that working-state with `task: "#9965 — verbose description"` triggers task-begin with number=9965. Carried in skill working-state since cycle 1334; safe to file now that #9965 has shipped to pending-test.

WallyDoodlez:
## Summary

Fix `cycle_pre._enforce_branch` parsing of the working-state `task` field so it correctly extracts the issue number from the verbose `#NNNN — description` convention (not just bare `#NNNN`).

Old code: `task.lstrip(\"#\").strip()` + `.isdigit()` — works for `#9965` and `9965` but silently no-ops for `#9965 — desc`, leaving the agent on the wrong branch.

New code: `re.match(r\"#*\s*(\d+)(?:\s|—|\$)\", task.lstrip())` with anchored capture.

Closes #10072.

## Behavior

| Input | Old | New |
|-------|-----|-----|
| `#9965` | accept → 9965 | accept → 9965 |
| `9965` | accept → 9965 | accept → 9965 |
| `#9965 — desc` | **reject** (regression) | accept → 9965 |
| `# 9965` | accept → 9965 | accept → 9965 |
| `##4942` | accept → 4942 | accept → 4942 |
| `9965-fix` | reject | reject |
| `#abc` | reject | reject |

## Tests

`tests/test_cycle_pre.py::TestEnforceBranch` — 19 tests, all passing:
- 11 positive parametrized cases (bare, hash, verbose em-dash, whitespace variants, multi-hash typos, ASCII-dash separator)
- 5 negative parametrized cases (`no-number-here`, `#`, `#abc`, `9965-fix`, `9965abc`)
- 3 pre-existing tests (calls task-begin, switches working branch, stays on working branch)

## External review iterations

| Round | Reviewer | Findings | Status |
|-------|----------|----------|--------|
| 1 | DeepSeek | 3 warnings (whitespace regression, missing whitespace tests, relaxed-match) | All fixed in `5851b7f1` |
| 2 | DeepSeek | 1 warning (`#?` rejected `##NNNN` typos) | Fixed in `427b4e20` |
| 3 | Claude (DS output too short, auto-fallback) | NO_FINDINGS | Converged |

## Test plan

- [ ] CI runs `python tests/run_tests.py` — should be green (4 unrelated pre-existing ERRORs in `test_deterministic_qa_framework::TestVerificationSubSkill` due to missing fixture `references/sub-skills/roles/qa/verification.md` — not in scope of this change).
- [ ] Verify the fix resolves the production symptom: after merge, a feature task with verbose working-state form should trigger `git_ops.py task-begin` automatically (no manual checkout needed; the bug memory rule `feedback_always_checkout_feature_branch` can be retired).

### DIFF
--- a/references/scripts/cycle_pre.py
+++ b/references/scripts/cycle_pre.py
@@ -210,7 +210,7 @@ def _enforce_branch(role, working_state):
         # bare ("#9965", "9965"), verbose ("#9965 — description"), and
         # whitespace-tolerant ("# 9965") forms. Trailing anchor rejects
         # glued suffixes like "9965-fix" that old `.isdigit()` also rejected.
-        m = re.match(r"#?\s*(\d+)(?:\s|—|$)", task.lstrip())
+        m = re.match(r"#*\s*(\d+)(?:\s|—|$)", task.lstrip())
         if m:
             number = m.group(1)
             result = _run_script("git_ops.py", "task-begin", role, number)
--- a/references/scripts/cycle_pre.py
+++ b/references/scripts/cycle_pre.py
@@ -20,6 +20,7 @@ Exit codes:
 """
 
 import json
+import re
 import shlex
 import subprocess
 import sys
@@ -205,9 +206,13 @@ def _enforce_branch(role, working_state):
     status = working_state.get("status", "none")
 
     if task != "none" and status == "in-progress":
-        # Extract issue number from task field (e.g. "#4942" -> "4942")
-        number = task.lstrip("#").strip()
-        if number.isdigit():
+        # Extract issue number from task field (#10072). Anchored to match
+        # bare ("#9965", "9965"), verbose ("#9965 — description"), and
+        # whitespace-tolerant ("# 9965") forms. Trailing anchor rejects
+        # glued suffixes like "9965-fix" that old `.isdigit()` also rejected.
+        m = re.match(r"#?\s*(\d+)(?:\s|—|$)", task.lstrip())
+        if m:
+            number = m.group(1)
             result = _run_script("git_ops.py", "task-begin", role, number)
             if result.returncode != 0:
                 # Non-fatal — log and continue on current branch
--- a/tests/test_cycle_pre.py
+++ b/tests/test_cycle_pre.py
@@ -1221,6 +1221,70 @@ class TestEnforceBranch:
         assert len(task_begin_calls) == 1
         assert "4942" in task_begin_calls[0]
 
+    @pytest.mark.parametrize("task_field,expected_number", [
+        ("#9965", "9965"),
+        ("9965", "9965"),
+        ("#9965 — 6274.2 AC2.8 catch-up (PM option-3 path) [PAUSED]", "9965"),
+        ("#10072 — verbose description with em-dash", "10072"),
+        ("9965 — bare-number verbose form", "9965"),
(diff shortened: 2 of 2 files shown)


WallyDoodlez:
## QA Results

**Status**: PASS

Regex extraction at `references/scripts/cycle_pre.py:208-213` correctly parses bare, verbose, em-dash, and whitespace-padded task forms while rejecting glued suffixes (`9965-fix`, `9965abc`). DS-review fixup commit 427b4e20 also preserved multi-hash tolerance (`##4942`, `###9965 —`).

Verification:
- `tests/test_cycle_pre.py::TestEnforceBranch`: 19/19 pass (11 positive + 5 negative parametrized + 3 existing).
- Full `python tests/run_tests.py`: 2529 PASSED, 0 FAILED, 6 SKIPPED, 8 ERROR.

The 8 errors are **all** pre-existing #9965 leftovers (not caused by #10072):
- 4× `test_event_mode_fragments::TestAc6M62ManifestWiring` — hardcodes `ROLES = ['dev', 'pm', 'qa', 'dm']`, looks for `references/roles/dev/includes-events.yml` (renamed to `worker` post-#6274.2).
- 4× `test_deterministic_qa_framework::TestVerificationSubSkill` — looks for `references/sub-skills/roles/qa/verification.md` (renamed to `verifier`).

Both pre-existing on main per `git checkout origin/main` comparison. Filed as **#10156** for skill follow-up. Skill correctly flagged the `test_deterministic_qa_framework` set as unrelated; the `test_event_mode_fragments` set wasn't noticed but has identical root cause.

Auto-merging via harness.

PR #10379, opened 2026-05-30
seed: Agent Skill Dev Team preset (L1-L3) + L4 project-local from session memory

WallyDoodlez:
## Summary

- Updates `references/presets/software-dev/manifest.yaml`: renamed display to **"Agent Skill Dev Team"** with a precise description framing this as SquidSquad's self-development preset
- Creates 4 new-format L4 files in `.squidsquad/project/`: `pm.md`, `worker.md`, `verifier.md`, `dm.md` — each with H2 sections (Identity / Soul / Instructions / Project Context) per `docs/COMPOSE-ARCHITECTURE.md` §5
- Content distilled from existing old-format L4 (kept as-is for current-runtime use), this session's accumulated user-memory feedback rules, and project memories

## L4 content sources (per role-class)

| Role | Soul drawn from | Instructions drawn from | Memory rules folded in |
|------|-----------------|------------------------|-----------------------|
| **PM** | `pm-soul-directives.md` | `pm-instructions.md` + `shared-instructions.md` | pm_docs_only, trust_script_output, scan_approval_tiers, pm_scan_process_only, fix_pm_bugs_immediately, bugs_behavior_only, trd_prd_delivery_model |
| **Worker** | `worker-soul-directives.md` | `worker-instructions.md` + shared | pm_docs_only (the code-owner side), comprehension_tests_required, skill_sonnet_subagents, clone_isolation, ds_review_per_change |
| **Verifier** | `verifier-soul-directives.md` | `verifier-instructions.md` + shared | no_ship_with_gaps, no_ship_failed_tc, dont_ask_before_verifying, comprehension_testing, test_workflow_separation, preserve_qa_tests, auto_approve_bugs |
| **DM** | `dm-soul-directives.md` | `dm-instructions.md` + shared | dm_optional, dm_sonnet_subagents, chat_subskills_deferred, harness_vision |

## Project Context content (in each role's file)

- **Project**: SquidSquad — multi-agent dev framework
- **Domain**: Claude agent / skill development
- **Stack**: Python 3.10+, Markdown, GitHub Issues, gh CLI
- **Repo**: https://github.com/WallyDoodlez/SquidSquad
- **Phase**: TRD-polish (2026-05-30) — PRD/task-shaping comes next
- **Self-hosting**: SquidSquad uses SquidSquad to build SquidSquad
- Plus role-specific recent context

## Architecture compliance

[...]

PR #10378, opened 2026-05-30
docs: INSTALLER-ARCH §10 — rewrite upgrade flow as sequential L4/config migration (#10376)

WallyDoodlez:
## Summary

Closes #10376. Rewrites INSTALLER-ARCH §10 (Upgrade flow) from the "detect → confirm changeset → pull → recompose → restart" model to the cleaner **upgrade-as-sequential-migration** model.

## Why

The previous model had four architectural gaps surfaced by DS audits + the cycle-2026-05-30 design discussion:

- "Latest upstream version from source release metadata" was unspecified (#10375 had been filed to design a `version.py` helper — now obsolete)
- Changeset summary required diffing two source-tree versions with no defined mechanism
- L4 / config schema changes had no migration path (the "may add new fields with defaults" caveat only handled additive changes)
- Install and upgrade were different paths with different semantics — "idempotent re-runs" advertised in §2.3 didn't really hold

## The new model

**Upgrade = sequential migration.** Every release that breaks L4 or `.squidsquad/config.md` schema ships a per-version migration file at `references/migrations/v<N-1>-to-v<N>.md`. The upgrade flow walks them in order against the operator's current L4 + config:

1. Pull source updates
2. Read installed version from `.squidsquad/config.md` `squidsquad_version:` (installer's own version from `references/VERSION`)
3. **Walk migrations in order**, each step gated by the same three-gate model as `l4-curation` (DS audit → mini-CQ → compose dry-run)
4. `compose.py deploy-all` regenerates CLAUDE.md outputs
5. Restart agents via harness HTTP

## New subsections under §10

| § | Content |
|---|---|
| §10.1 | Migration file format (prose, mechanical + judgment-call examples) |
| §10.2 | What stays untouched during upgrade (scoped to upgrade context after audit fix) |
| §10.3 | What's regenerated |
| §10.4 | Edge cases — missing version stamp, missing migration files, partial-walk safety, DS reject, dry-run failure, skipped versions |
| §10.5 | Framework-shipped vs operator-written |

## Commits

| SHA | Scope |
|---|---|
| ` 947c0ba8 ` | §10 rewrite (full replacement of upgrade-flow spec) |
| ` 7736c82a ` | DS internal audit M2 fix — §10.2 "untouched" invariants scoped to upgrade context |

## Audit trail

DS internal audit run on the rewrite. 5 findings:
- 1 actionable (M2 — §10.2 "never bulk-overwritten" was absolute, scoped to upgrade-only in ` 7736c82a `)
[...]

PR #10380, opened 2026-05-30
docs: post-merge audit fixes — 3 COMPOSE-ARCHITECTURE internal findings

WallyDoodlez:
## Summary

Post-merge audit (after #10378 + #10379 both landed) on the 5 architecture docs identified 3 genuinely merge-introduced issues in `COMPOSE-ARCHITECTURE.md`. This PR fixes them. Other cross-pair findings are recycled pre-existing wording flagged in earlier audit rounds — out of scope here.

## Fixes

- **§3.3 identity row** — the merged identity-row text combined two PRs' wording into something that read as contradictory (\"append only\" vs \"cannot insert between Identity prose and the Boundaries list\"). Rewrote to make the two rules reinforcing: identity is append-only, so there's no insert-before/insert-after to begin with — Boundaries structurally cannot be split or reordered by L4. Added compose-side guarantee that L1 Boundaries is always the last L1-L3 fragment.
- **§1 path-keying table** — config.md was described in prose as a sibling of project/ and <alias>/ but never appeared in the path-keying table. Added a 4th row.
- **§4.6 atomic-triple vs Assemble: no opt-out** — the failure-mode prose said the artifact triple is atomic, but the Assemble: no opt-out emits only CLAUDE.md. Clarified: under default Assemble: yes the triple is atomic; under Assemble: no the triple guarantee doesn't apply.

## Test plan

- [ ] Confirm §3.3 identity row reads as internally consistent
- [ ] Confirm §1 path-keying table now has 4 rows including config.md
- [ ] Confirm §4.6 failure-mode prose distinguishes Assemble: yes (atomic triple) from Assemble: no (CLAUDE.md only)


WallyDoodlez:
**Closing as superseded.**

Post-merge audit fixes filed yesterday have been overtaken by later architectural simplifications this session:

- **Fix 1 (§3.3 identity row)**: ✅ already addressed on main via a later edit. The current line reads internally consistent.
- **Fix 2 (§1 path-keying table adds config.md row)**: references the `event-driven:` config flag which was **dropped** (see PR #10382). The fix would re-introduce stale terminology. If a config.md row in the path-keying table is still wanted, file as a fresh small PR with current terminology.
- **Fix 3 (§4.6 atomic-triple vs `Assemble: no`)**: the `Assemble:` opt-out itself was **dropped** this session (see commit on PR #10383 + memory rule `project_assemble_unconditional`). The whole conditional triple-atomicity branch is moot — assemble is unconditional; the triple is always emitted atomically.

Not re-doing the salvageable bits in this PR because the diff fights with the current state. If §1 path-keying table needs config.md, file fresh.

Issue #9588, opened 2026-05-20
TASK: Lazy-load mode-specific instructions at boot via config check, instead of inlining at compose time


WallyDoodlez:
**Reported By**: pm-lead
**Priority**: Medium

**Reported By**: pm-lead (human direction, cycle 1535)
**Priority**: Medium
**Supersedes**: #9580 (incorporates its scope + broader architecture change)

## Goal

Today, `compose.py` reads the `event-driven` config flag at deploy time and inlines either polling-mode (`ralph-loop-overview.md`) OR event-mode (`common-events/l1-base.md` + 5 others) fragments into the composed CLAUDE.md. Result: large composed file with the wrong half potentially inlined.

Switch to a **boot-time lazy-load model**: compose.py emits a small bootstrap snippet that tells the agent to read `config.md` on startup, then Read the appropriate fragment based on event-driven + harness-reachability state.

## Why now (not deferred to event-mode flip)

- Polling mode is the current default. Polling agents today carry the full polling instruction set inline — fine, but the architecture should be uniform.
- Event-mode fallback to polling (per #9580) is already going to need 'read polling fragment on demand' — so the mechanism is being built anyway. Generalize it once instead of doing it twice.
- Smaller composed CLAUDE.md across the board (only the mode-specific instructions move out — agent identity, role L2, project L4 still inline).

## Proposed mechanism

`compose.py` emits a boot bootstrap section in every role's CLAUDE.md, replacing the current mode-specific inline includes:

```
## Boot — Mode Detection (Lazy Load)

On first invocation, do this BEFORE any other action:

1. Read `.squidsquad/config.md` to determine the active wake mode:
   - If `event-driven: yes` → POLL harness at `.squidsquad/.harness-port`. On reachable, Read `references/sub-skills/common-events/l1-base.md` and follow its boot sequence. On unreachable, fall through to step 2.
   - If `event-driven: no` OR fallthrough from above → Read `references/sub-skills/common/ralph-loop-overview.md` and follow its instructions (which include invoking `/loop 30m execute one Ralph Loop cycle`).

2. The loaded instructions become your active wake-mode contract.

3. Re-check mode at each cycle boundary (`cycle_post.py` already loads config — extend it to emit a `reload_mode` signal on flag change).
```

## What stays compose-time inline

- L1 agent foundation (identity, tracker protocol, git protocol — mode-agnostic).
- L2 role responsibilities.
- L3 domain variant.
[...]

Issue #9580, opened 2026-05-20
TASK: Event-mode degraded fallback should be /loop polling, not bespoke sleep+retry


WallyDoodlez:
**Reported By**: pm-lead
**Priority**: Medium

**Reported By**: pm-lead (per human direction, cycle 1535)
**Priority**: Medium

## Current state (common-events/l1-base.md §3, in-stream gap branch)

When the harness is unreachable at boot or becomes unreachable post-bootup, the spec puts the agent into a bespoke 'degraded mode': work directly from the forge via `work_queue()`, sleep 60s on empty queue, retry `bootup-complete` POST with backoff capped at 5 min. **Do NOT enter improvement-scan cool-down.** Custom loop, only used in this one degraded scenario.

## The problem

That degraded mode is novel + unproven. The session of 2026-05-19→20 demonstrated the harness can wedge in multiple ways (two distinct ProactorEventLoop bugs across cycles 1500–1527). Until we're confident the harness is durable and has reliable recovery, **the agents should NOT be running a bespoke fallback that has its own failure modes** — they should fall back to the battle-tested `/loop 30m execute one Ralph Loop cycle` polling mode that's served us for years.

The polling mode:
- Has been used continuously through every harness outage in this session.
- Recovers automatically — if anything goes wrong, the next /loop fire picks up where we left off.
- Doesn't accumulate state that gets lost on agent restart.
- Has well-understood cadence (config-driven, default 30 min).

The bespoke degraded mode:
- 60s sleep is hard-coded, not config-driven.
- Skips the improvement-scan cool-down (which is the cycle's quiet-work mechanism).
- Has its own subtle exit condition (`bootup-complete` success) that the polling-mode boot path doesn't share.
- Hasn't been exercised end-to-end yet (no integration test — it's covered by §4.5 in #9398's deferred scope).

## Proposed change

Rewrite `common-events/l1-base.md` §3 unreachable branch to:

1. Agent invokes `/loop 30m execute one Ralph Loop cycle` (or whatever interval comes from `config.md`).
2. Polling-mode Ralph Loop drives the agent (same instructions we ship today in `ralph-loop-overview.md`).
3. `cycle_pre.py` checks harness reachability each cycle. When reachable AND we have a cursor, agent: (a) cancels the /loop schedule, (b) skims events from cursor forward, (c) emits `bootup-complete`, (d) invokes Monitor tool with `event_poll.py`. Same step-5-finish as Case A.
[...]

PR #10382, opened 2026-05-30
docs: drop event-driven: config flag; boot-probe selects wake mode

WallyDoodlez:
## Summary

Architectural correction confirmed 2026-05-30: **event mode is the unconditional architecture; loop is the automatic boot-time fall-back when the harness is unreachable.** There is no `event-driven:` config field, no compose-time manifest gate, no operator-flipped mode flag.

- Selection happens **at boot only** via a 5-second HTTP probe of the harness.
- Bound per-session; the next session restart re-probes.
- Mid-session bus failures degrade individual cycles to tracker reads (per AGENT-RUNTIME §4.5) but do not flip the wake mechanism.
- The historical \`#9580\` / \`#9588\` "no automatic runtime fall-back" decision applied specifically to mid-session mode-flipping (which is still rejected); boot-time fall-back is canonical.

## Files touched

- \`docs/AGENT-RUNTIME.md\` — §2 / §4 / §8 rewritten; new rev 11 entry maps prior rev 6/7/9 framings as historical
- \`docs/COMPOSE-ARCHITECTURE.md\` — §3.0 / §3.2 / §4.4 / §5.1.1 / §5.7 / §6.5 reflect mode-agnostic compose; \`includes-events.yml\` split retired
- \`docs/INSTALLER-ARCH.md\` — \`event-driven:\` dropped from §3.2 outputs, §4.5 spec example, §6 \`event_poll.py\` note, §10 migration example
- \`docs/sub-skill-catalog.md\` — \`common-events/\` reframed as "all roles" with runtime bus-failure fall-back paths

## Memory

Saved \`project_event_mode_default.md\` so future audits / dev tasks reject any reintroduction of the field.

## Test plan

- [ ] Verify no production-spec text references \`event-driven:\` as a live field (archive/ entries are historical and stay)
- [ ] Confirm the new §8.3 boot decision tree is self-consistent with §4.5 runtime fallback semantics
- [ ] Spot-check §5.7 worked example flows for both event-mode and loop-mode-fallback runtime cases

## Code follow-ups (out of scope, file separately)

- \`compose.py\`: drop \`_load_manifest\`'s wake-mode branch + \`config.get_wake_mode()\`; remove \`includes-events.yml\` lookups
- \`thin_launcher.py\` (or successor per HARNESS §14): implement boot-time harness probe + wake-mode bind
- \`config.py\`: drop \`event-driven\` getter
- Per-role \`includes-events.yml\` files: fold their content into the unified \`includes.yml\`
- Cycle-body Python helpers: ensure bus-read paths catch connection errors and fall through to tracker reads


WallyDoodlez:
## Added: drop upgrade-flow framing (commit f7dfe453)

Second architectural simplification in the same arc as the first commit:

- There is no distinct "upgrade flow". Re-running the installer with `.squidsquad/` already present runs the same phases as a fresh install.
- The existing-install case is handled by a **migration-walk step** (§10) that consults per-version `references/migrations/v<N-1>-to-v<N>.md` files — same mechanics as before, but now one step of the standard flow rather than a parallel top-level flow.
- §10.2 "What is preserved during upgrade" deleted; replaced with the simpler rule: the installer never wipes existing content; migrations are the only mutator.
- Resolves the pre-existing INSTALLER M1 audit finding (§10.2 vs §11.2 contradiction on config.md preservation).

Memory rule saved: `project_upgrade_is_fresh_install.md`.

Both commits share the same pattern: collapse dual-path framings into one canonical path with auto-detection.

Issue #10348, opened 2026-05-26
ISSUE: health_check._read_interval catches Exception subclasses but not SystemExit from config.get_field — uncaught sys.exit(1) on missing interval field


WallyDoodlez:
**Reported By**: skill-lead
**Severity**: Low

**Reported By**: skill-lead
**Severity**: low
**Found by**: skill-lead (improvement-scan, cycle 1435)
**File**: references/scripts/health_check.py (L61-75)

**Finding**: `_read_interval` wraps the config lookup in `try / except (ImportError, ValueError, TypeError): pass` and falls through to a hard-coded `return 30` default. But `config.get_field` calls `sys.exit(1)` (raises `SystemExit`) when the requested field is absent from `config.md` and not in `_FIELD_DEFAULTS`. `_FIELD_DEFAULTS` only contains `{"event-driven": "no"}` — there is no default for `interval`, so a config.md that lacks `Iteration Interval > Minutes` triggers `sys.exit(1)` inside `get_field`. Same path fires if `_read_config()` itself `sys.exit(1)`s on a missing config.md (line 121 in config.py).

`SystemExit` is a `BaseException`, not an `Exception`, so the existing catch list does NOT catch it. The script aborts with exit code 1 and an "ERROR: Field 'interval' not found" line on stderr — **the documented behavior of returning the 30-minute default never fires**. Exit 1 is also the "stalled or unknown" code per the module docstring, so callers parsing the exit code see a misleading "agent is stalled" report when the real failure is config absence.

**Concrete consequence**: On a fresh checkout or partially-initialized clone where `.squidsquad/config.md` lacks the interval field (rare in practice but possible during upgrade/migration windows), `python references/scripts/health_check.py` exits 1 silently. The PM Step-9242 manual stall-recovery path that shells out to health_check (`boot_remote` callers) sees the wrong signal.

**Same defect-family**: `cycle_post._config_get` already documents and handles this exact case (`except BaseException:` per the comment about config.get_field calling sys.exit(1)). `health_check._read_interval` missed the lesson.

**Recommendation**: widen the catch to `except (BaseException,)` or `except (SystemExit, ImportError, ValueError, TypeError):` so the function falls through to its documented 30-minute default on **any** config-read failure. Mirrors the cycle_post._config_get pattern.

[...]

PR #10424, opened 2026-05-31
skill: #10348 — health_check._read_interval catches SystemExit + drops dead imports

WallyDoodlez:
Closes ​#10348

### Summary

`config.get_field` calls `sys.exit(1)` (raising `SystemExit`, a `BaseException`) when the requested field is absent from `config.md`. `health_check._read_interval` previously caught only `(ImportError, ValueError, TypeError)`, so the exit propagated up and aborted the script — the documented 30-minute fallback never fired.

This widens the catch to `(SystemExit, Exception)`: covers the bug case plus every standard `Exception` subclass, while letting `KeyboardInterrupt` and `GeneratorExit` propagate so Ctrl+C still aborts. Per DS-review iter1 feedback, this is narrower than the `except BaseException` precedent in `cycle_post._config_get` (which would silently swallow Ctrl+C).

Also drops three dead imports (`os`, `platform`, `subprocess`) from the issue body's 'out of scope but observed' cleanup.

### Acceptance Criteria

- [x] `_read_interval` returns the 30-minute default when `config.get_field` raises `SystemExit` (regression test `test_system_exit_returns_default`)
- [x] `KeyboardInterrupt` continues to propagate (regression test `test_keyboard_interrupt_propagates`)
- [x] Pre-existing catches for `ImportError`, `ValueError`, `TypeError` still fall through to the default (existing tests unchanged)
- [x] Dead imports `os` / `platform` / `subprocess` removed

### Changes

- **Files**: `references/scripts/health_check.py` (4-line change in `_read_interval`, 3 lines removed from imports, docstring updated); `tests/test_health_check.py` (+2 tests in `TestReadInterval`)
- **What**: One catch clause widened to `(SystemExit, Exception)`, docstring updated to call out deliberate `KeyboardInterrupt` propagation, three unused imports deleted.
- **Why**: Restore the documented 30-minute fallback behavior on missing `Iteration Interval` field; mirror the cycle_post precedent but narrower to avoid the Ctrl+C-eating side effect.

### Verifier Status

- [x] Unit tests passing (41/41 in `test_health_check.py`)
- [x] Acceptance criteria met
- [x] External code review converged: iter1 (DeepSeek) flagged BaseException-too-broad → iter2 (Claude subagent fallback per model_router auto-fallback rule) NO_FINDINGS

Full-suite errors on this branch (`test_agent_boundaries`, `test_manifest`, `test_feat_6126_harness_merge`, one transient `test_04_shipped_and_closed` gh-API flake) reproduce on `main` without these changes — pre-existing, not caused by this commit.

### DIFF
--- a/references/scripts/health_check.py
+++ b/references/scripts/health_check.py
@@ -23,10 +23,7 @@ Exit codes:
 
 import io
 import json
-import os
-import platform
 import re
-import subprocess
 import sys
 import time
 from pathlib import Path
@@ -61,8 +58,12 @@ HEALTH_EMOJI = {
 def _read_interval():
     """Read the iteration interval from config.md. Default 30.
 
-    Uses config.py get_field (section-aware via FIELD_MAP) to avoid
-    matching unrelated 'Minutes' occurrences in other config sections.
+    Catches `SystemExit` because `config.get_field` calls `sys.exit(1)`
+    when the field is absent and not in `_FIELD_DEFAULTS`, same for
+    `_read_config()` on missing config.md. SystemExit is a BaseException,
+    not an Exception, so a narrow `except Exception` would miss it (#10348).
+    KeyboardInterrupt deliberately propagates — Ctrl+C must abort, not
+    silently fall through to the default.
     """
     try:
         sys.path.insert(0, str(SCRIPT_DIR))
@@ -70,7 +71,7 @@ def _read_interval():
         val = get_field("interval")
         if val:
             return int(val)
-    except (ImportError, ValueError, TypeError):
+    except (SystemExit, Exception):
         pass
     return 30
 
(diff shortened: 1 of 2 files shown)


WallyDoodlez:
## Verifier Results

**Status**: PASS
**Test Plan**: .squidsquad/qa/planning/TEST-PLAN-10348.md (Verifier-owned, derived from issue body — not from the PR diff)
**Verifier TCs**: 6/6 pass (`pytest .squidsquad/qa/planning/TEST-10348-tests.py`)
**Worker regression suite**: 41/41 pass (`pytest tests/test_health_check.py`)

All ACs observably satisfied:
- AC-1 SystemExit → fall-through to 30 — verified both at unit level (TC-1) and via a live subprocess against a real config.md missing the interval field (TC-2 — the documented pre-fix bug no longer reproduces)
- AC-2 ImportError/ValueError/TypeError still return 30 (no regression)
- AC-3 KeyboardInterrupt propagates — drift guard against future widening to BaseException
- AC-4 Dead imports (os, platform, subprocess) dropped — AST-walk confirmed

Verifier regression suite promoted to `tests/test_feat_10348_verifier.py`.

Issue #10657, opened 2026-06-01
TASK: PRD-C / Story C8: Conflict pre-emption (reads linked composite + reframes)


WallyDoodlez:
**Reported By**: pm-lead
**Priority**: Medium

**Parent PRD:** [[compose-l4-customization]] §7.1 + §4.6 pairing (story C8 row — conflict pre-emption)

## Goal

Before drafting an L4 op, `l4-curation` reads the LINKED composite for the target slot and detects whether the new entry would materially contradict existing L1-L3 prose. If so: prefers `replace` reframing, rewords to refine, or surfaces to human — never silently writes the conflicting op.

## Acceptance criteria

1. After the decision tree classifies an `insert-*` or `append` op, `l4-curation` reads the linked composite for `(slot, role-class)` and runs a conflict check.
2. Conflict detection reuses the vocabulary established by PRD-B B4 (#10445, shipped) — "materially contradicting prose between layers". This story does NOT re-implement contradiction detection; it invokes B4's detector or a B4-shaped LLM call.
3. On contradiction detected: agent surfaces to human with quotes from both sides + reframe options (replace-reframe / abandon).
4. On no contradiction: proceeds to Gates 1-3 normally.
5. Tests: at least (a) clean insert → no surfacing, (b) inserting a constraint that contradicts an L2 instruction → surface + reframe offered, (c) `replace` op never triggers conflict pre-emption (replace by definition supersedes).

## v1 coexistence

Pre-emption logic on the v2 invocation path only.

## Dependencies

- C1 (sub-skill prose) for integration point
- B4 (#10445) — shipped ✓ (provides the conflict-detection vocabulary)

## Effort

M (logic + integration + tests; the LLM call is the heavy operation).

WallyDoodlez:
**pm-lead**: Approving per operator direction 2026-06-01 (PRD-C batch).

- All 10 PRD-C stories approved together; skill picks per the PRD's recommended pickup order
- C1 (sub-skill prose) + C10 (comprehension tests) recommended to run in parallel as the bootstrapping pair
- Then C2 (wire), C4 (mini-CQ), C6 (atomic write), C3 (DS audit), C5 (dry-run via shipped A4.5), C7 (recovery), C8 (conflict pre-emption via shipped B4), C9 (counter-entry)
- §9a coexistence: all stories ship without touching v1; behavior changes only at PRD-E E6 cutover

WallyDoodlez:
[...]

Issue #10656, opened 2026-06-01
TASK: PRD-C / Story C7: Recompose-failure recovery (git revert + alert)


WallyDoodlez:
**Reported By**: pm-lead
**Priority**: Medium

**Parent PRD:** [[compose-l4-customization]] §7.4 step 3 race (story C7 row — failure recovery)

## Goal

Handle the post-commit recompose-failure race: if `compose.py deploy-all` (triggered by harness file-watch per PRD-E) fails after C6's commit lands, revert the L4 commit and alert.

## Acceptance criteria

1. C6 emits a known event (or polls for harness's recompose result) so the agent knows the post-commit recompose succeeded or failed within a bounded window.
2. On recompose success: log success in iteration file, continue cycle normally.
3. On recompose failure (or timeout): `git revert <commit-sha>` the L4 commit, push the revert, then comment on the relevant tracker issue (or surface inline in conversation) with the recompose diagnostic.
4. The revert must NOT itself be a rebase — use a `git revert` commit (additive history).
5. Iteration log records: (a) the original L4 write commit SHA, (b) the recompose failure reason, (c) the revert commit SHA.
6. Tests: at least (a) recompose succeeds → no revert, log success; (b) recompose fails → revert lands + alert; (c) recompose times out → same as fail.

## v1 coexistence

Same as C6 — the L4 file lives outside v1's compose path; revert leaves no v1 effect.

## Dependencies

- C6 (atomic write) before C7.
- PRD-E harness file-watch contract for the recompose trigger (C7's contract is "we wait for recompose result"; the trigger mechanism is PRD-E's lane).

## Effort

S (pure failure-path handling).

WallyDoodlez:
**pm-lead**: Approving per operator direction 2026-06-01 (PRD-C batch).

- All 10 PRD-C stories approved together; skill picks per the PRD's recommended pickup order
- C1 (sub-skill prose) + C10 (comprehension tests) recommended to run in parallel as the bootstrapping pair
- Then C2 (wire), C4 (mini-CQ), C6 (atomic write), C3 (DS audit), C5 (dry-run via shipped A4.5), C7 (recovery), C8 (conflict pre-emption via shipped B4), C9 (counter-entry)
- §9a coexistence: all stories ship without touching v1; behavior changes only at PRD-E E6 cutover

WallyDoodlez:
**skill-lead**: Picking up. Status → In Progress.

WallyDoodlez:
[...]

Issue #8950, opened 2026-05-18
TASK: defense-in-depth: code-review loop, QA, DM each check planning artifact against implementation


WallyDoodlez:
**Reported By**: pm-lead
**Priority**: High

> **AUTHORITATIVE SCOPE: `.squidsquad/pm/planning/TEST-PLAN-8950.md` (entire file). Read that artifact in full. The bullets below are a summary; the planning artifact is the contract.**

---

## Defense-in-depth (Audit C Finding 4, incident response 2026-05-18)

The 2026-05-18 incident showed all 4 implementation gates (skill / code-review / QA / DM) keyed off implementation output, none checked planning contract. #8916 + #8917 address gate #1 + root cause. This ticket adds defense to gates #2/#3/#4.

## Scope summary (TEST-PLAN-8950.md is the contract)

Three sub-skill edits via the compose stack:

- **Gate #2 — Code-review loop §9c**: feed CONTEXT.md / TEST-PLAN as `--input-files`; prompt deepseek to check architectural locks. (Coordinates with #8916.)
- **Gate #3 — QA AC walk**: before `pending-test → pending-ship`, walk TEST-PLAN-N.md AC list and confirm each AC is observably satisfied. Tests-passing necessary but not sufficient.
- **Gate #4 — DM contract-citation soft gate**: PR must reference the locked planning artifact; else block + route back to QA.

## Acceptance

See `.squidsquad/pm/planning/TEST-PLAN-8950.md` for AC-1..AC-4 and CQ-1..CQ-3.

## Coordination

§9c overlap with #8916 — recommend single PR if both approved together; else whichever ships first defines the fragment.

WallyDoodlez:
**pm-lead**: Human approved for execution. Status -> approved. Planning artifact: `.squidsquad/pm/planning/TEST-PLAN-8950.md` (deepseek NO_FINDINGS). Skill: read the planning artifact before implementing per the AUTHORITATIVE SCOPE banner.

WallyDoodlez:
**skill-lead**: Picking up — TEST-PLAN-8950.md is the contract. Three sub-skill edits (dev §9c, qa verification, dm delivery-packaging) + compose.py deploy-all + AC-4 PM CLAUDE.md byte-identical check. Working on branch squidsquad/task/8950. Status -> In Progress.

WallyDoodlez:
[...]

PR #10667, opened 2026-06-01
skill: feat(l4): C9 removal flow — counter-op (default) + in-place delete (#10658)

### LINKED ISSUE #10658: TASK: PRD-C / Story C9: Counter-entry / removal flow (undo a prior customization)
**Reported By**: pm-lead
**Priority**: Medium

**Parent PRD:** [[compose-l4-customization]] §7.5 + §7.7 (story C9 row — counter-entry / removal)

## Goal

Implement the "undo a prior customization" dialog branch: produces either a counter-op (e.g., `replace` with empty body) or, with explicit human confirmation, deletes the targeted H3 block in-place.

## Acceptance criteria

1. `l4-curation` detects removal-shaped requests ("undo the incidents-check thing", "drop the last L4 entry for worker", etc.).
2. Removal dialog surfaces the targeted entry: quotes the existing L4 H3 block + commit SHA + authored-by metadata.
3. Two paths offered:
   - **Counter-op** (default): adds an op that cancels the prior behavior (e.g., `### replace step:cycle/file-bug` with an empty body — interpreted as a no-op replacement). Additive history, prior entry stays visible in git.
   - **In-place delete** (requires explicit human confirmation): rewrites the L4 file to remove the H3 block. The git commit shows the deletion in the diff (no history rewrite — additive in git terms, even though the file lost content).
4. Both paths go through Gates 1-3 (C3, C4, C5) and use the same C6 atomic commit flow.
5. Tests: at least (a) counter-op path → file gets new op, prior op preserved, (b) in-place delete with explicit confirmation → H3 block removed, (c) attempted in-place delete without explicit confirmation → falls back to counter-op offer.

## v1 coexistence

Same as C1-C7 — all on the v2 path.

## Dependencies

- C3, C4, C5, C6 (all gates + atomic write) before C9.

## Effort

S (dialog branch + counter-op generation; reuses existing gates and write).

WallyDoodlez:
Closes #10658.

## Summary
- Adds `references/scripts/l4_removal.py` — PRD-C / Story C9 helper that produces staged content for the existing gates 1-3 + C6 pipeline when the human asks to undo a prior L4 customization.
- 44 unit tests covering all 5 ACs including the three required AC5 paths (counter-op, in-place delete with explicit confirmation, fallback when in-place delete isn't confirmed) plus 6 edge cases (detection clause-boundary anchoring, ambiguous targeting, blame_lookup_fn exception swallowing, unparseable-L4 backoff, etc.).
- Enriches `references/sub-skills/common/l4-curation.md` with a new "Removal flow" section that documents the two paths, the detection/targeting/preview phases, and the failure-mode plumbing.

## Planning contract (per #8950 Gate #4)
- **Module shape**: ~340 LOC; 6 public functions (`is_removal_request` / `find_target_entry` / `format_target_preview` / `build_counter_op` / `build_in_place_delete` / `plan_removal`); 2 dataclasses (`TargetEntry` / `RemovalPlan`).
- **Orchestrator never raises** — failure modes carry in `RemovalPlan.path_chosen` ∈ {counter-op | in-place-delete | ambiguous | not-found | no-counter-op-possible | not-a-removal-request}. Callers branch on the field rather than catching exceptions, mirroring C5's `DryrunResult` / C6's `WriteResult` shape.
- **Injection seam**: `blame_lookup_fn` lets tests stub git-blame without invoking git. Exceptions from the seam are swallowed so audit-trail population can never block the removal.
- **Files added**: `references/scripts/l4_removal.py`, `tests/test_l4_removal_c9.py`.
- **Files modified**: `tests/run_tests.py` (registers new module), `references/sub-skills/common/l4-curation.md` (Removal flow section).
- **Files NOT touched**: `compose.py`, `model_router.py`, any L1-L3 source, the other `l4_*.py` gate helpers (no API drift).
- **§9a impact**: zero — v1 byte-stability gate green (5/5).
- **Operator-rule compliance**: in-place delete is opt-in only via `in_place_delete_confirmed=True` — silent autonomous deletion is impossible by construction. Detection regex anchors at clause boundaries so mid-sentence "delete" / "drop" / "forget" don't trip the removal flow.
- **Tests**: 44/44 green (0.12s).
- **Follow-ups**: C7 (#10656 recompose-failure recovery), C8 (#10657 conflict pre-emption).

## AC mapping
| AC | Implementation | Tests |
|---|---|---|
| 1: detect removal-shaped requests | `is_removal_request` — sentence-anchored verb regex | 15 parametrized cases + non-removal anti-cases |
| 2: surface targeted entry + commit SHA + authored-by | `find_target_entry` + `format_target_preview` | 7 tests (confident match / ambiguous / not-found / blame seam) |
| 3a: counter-op default (`### replace step:cycle/X` empty body) | `build_counter_op` + `plan_removal` default path | 5 tests (step-targeted emit / non-step skip / counter-op happy path) |
| 3b: in-place delete with explicit confirmation | `build_in_place_delete` + `plan_removal(in_place_delete_confirmed=True)` | 5 tests (excision + reparseability + sibling preservation) |
| 4: both paths reuse gates 1-3 + C6 | helper produces staged TEXT, doesn't dispatch gates | (no test — module surface is the contract) |
| 5: tests for counter-op / in-place delete / fallback | covered above | 44 unit tests total |

## Test plan
- [x] `python -m pytest tests/test_l4_removal_c9.py -v` — 44/44 green
- [x] `python -m pytest tests/test_v1_byte_stability_9a.py -v` — 5/5 green (no v1 regression)


WallyDoodlez:
**Post-review fix-up (commit `86508cce`)**

Independent code review caught a real **BLOCKER**: my original counter-op design assumed `_apply_replace_step` with empty body would "fall back to L1-L3", but it literally blanks the step. Worse, even with that fallback added, the op processor is sequential — when the counter-op fires, the slot content already shows the customized body from the prior op, so a per-op "no-op" rule preserves the customization, not the L1-L3 body.

Right shape: **counter-op pair stripping** at the top of `apply_l4_ops`.

- `l4_op_processor._strip_counter_op_pairs(ops)` scans for `replace step:cycle/X` ops whose body is the `<!-- counter-op: ... -->` sentinel, pairs each with the most-recent prior `replace step:cycle/X` for the same step ID, and drops both before the per-op loop runs.
- Per-op semantics unchanged — a real empty-body `replace` still genuinely blanks the step (preserving the rare "intentionally clear this step" operator intent).
- `build_counter_op` restricted to `replace step:cycle/X` priors only. `insert-before`/`insert-after`/non-targeted ops route to `no-counter-op-possible` → in-place-delete (because pair-stripping can't help when the original op's content survives a subsequent replace).
- Added **TestCounterOpComposeIntegration**: composes `[original_replace, counter_op]` end-to-end through `apply_l4_ops` and asserts the L1-L3 body survives — the test that would have caught B1.

Also addressed CONCERNS C1-C4 + NITS N1-N2:
- **C1**: `_MIN_OVERLAP` drops to 1 when only one L4 op exists across the file.
- **C2**: detection regex extended with `kill` / `scrap` / `ditch` / `rip out` / `back out` / `get rid of` + `let's`/`let us` prefix anchor.
- **C3**: `find_target_entry` scoring includes H3 heading text so step-id tokens (`file-bug`) match.
- **C4**: in-place-delete H2/H3 boundary check uses regex anchors so `#### subhead` in a body doesn't break excision.
- **N1**: counter-op text ends with `\n`.
- **N2**: documented why `_normalize_slot` is duplicated.

Tests: 79/79 green (was 44 before; +35 from integration tests and op-processor coverage broadening).

Issue #11329, opened 2026-06-07
TASK: Runtime migration — per-event ack-cursor + working-state.md cursor cleanup (follow-up to #11328)


WallyDoodlez:
**Reported By**: pm-lead (split from #11328 — doc task is doc-only per operator clarification 2026-06-07)
**Priority**: High
**Depends on**: #11328 must ship first (canonical arch must be aligned before runtime migrates to match it)

## ⚠️ AC reframe ratified 2026-06-11 (PM)

The ACs as written below (drafted 2026-06-07) assumed **model A** (event_poll owns the cursor + acks at emit-time). Skill pickup analysis (2026-06-11) correctly identified that the canonical model post-#11328 D2 and polish-session is **model B** (event_poll emits bare `NUDGE\n`; agent owns acks). Model A breaks at-least-once delivery: if event_poll acked at emit-time, the cursor would advance before the agent processes the event, so an agent crash between emit and process silently loses the event. **Implementing the original AC text would re-create the exact doc-vs-runtime drift this task exists to kill.**

**Reframe ratified by PM (comment thread 2026-06-11)**. The reframe below supersedes the original ACs. The original AC text is preserved unchanged below this fence for audit-trail / archaeology.

### AC1 (reframed) — `event_poll.py` becomes nudge-only

- `event_poll.py` emits bare `NUDGE\n` (no payload, per AGENT-RUNTIME §8.0:773), drops ALL cursor logic
- New-event detection uses a local high-water-mark (NOT the harness cursor — that is the agent contract)
- Remove the pending-migration comment at line ~299
- Per-event ack semantic moves to the **agent** side (already implemented per `event-mode-contract.md:56,87,97`)

### AC2 (reframed) — `working-state.md` schema + transitional-note cleanup

- Remove the `- **Last Processed Event ID**:` line from the schema/templates (cursor now lives only in `.event-state.json`)
- Any code that previously read the cursor from `working-state.md` is removed or rerouted to `GET /events/cursor`
- **Retire the explicitly-tagged transitional notes written to be removed by #11329**: `cursor-management.md:14`, `event-mode-contract.md:50,98`, `instructions.md:186`

### AC3 (unchanged) — Harness ack consumer verification

- Per-event `ack-cursor` consumer verified at `harness.py:2018` → `advance_cursor(role, <single event_id>)` (line 973), regression-guarded by `find_positions` (line 858). **Dependency met.**
[...]

Issue #9415, opened 2026-05-20
TASK: Audit event id space — 32-bit hex prone to birthday collisions, consider ULID or 64-bit random


WallyDoodlez:
**Reported By**: pm-lead
**Priority**: Medium

**Reported By**: pm-lead (per Ilya0527/ALEF review on #8999 cycle 1511)
**Priority**: Medium

## Observed Behavior

`harness.py:1920` generates event ids as `os.urandom(4).hex()` — 8 hex chars / 32 bits / ~4.3B id space. Per the [birthday paradox](https://en.wikipedia.org/wiki/Birthday_problem), collision probability hits 50% around `sqrt(2 * 4.3B * ln 2)` ≈ 77k events. In practice, **collisions become non-negligible at ~65k events** in the bus lifetime.

This is **separate from and downstream of** #9265's verdict (which correctly dropped the in-stream gap scenario as architecturally inapplicable to the single-deque broadcast model). Id collisions create a different failure mode the broadcast model still permits:

- `EventStream.get_since(since_id)` (harness.py:475) does linear scan on `event.get("id") == since_id`. If two events share an id, the cursor anchors on whichever appears first, and the second one is silently skipped (or worse, the cursor advances back to before real work if the older duplicate's index is treated as authoritative).
- The eviction-gap path (#9331, just shipped) writes `oldest_id` into the agent's working-state cursor. If that id later collides with a newer event, the agent's cursor may resolve to the wrong position on next poll.

## Source attribution

Raised by Ilya0527/ALEF on #8999 comment 2026-05-20T02:33:53Z. ALEF-PAT-013 (their pattern catalog, https://n50.io/patterns/013). Their two other points in the same comment — §4.3 cooldown-write atomicity, §4.7 DM no-sub-loop assertion — are tracked separately.

## Scope

1. Decide on a wider id format. Two reasonable options (skill picks):
   - **ULID** (128-bit time-sortable, lexicographic ordering preserves emission order — useful for any future skim-then-advance logic).
   - **64-bit random hex** (`os.urandom(8).hex()` — collision-resistant past any practical event volume; doesn't add ordering).
2. Migration: change `_emit_event` (harness.py:1920) to the new generator. Existing `os.urandom(4).hex()` ids in retained events stay valid (no migration of the deque needed — agents process events one at a time regardless of width).
[...]

PR #11411, opened 2026-06-11
skill: #11403 — declare + provision harness runtime deps

WallyDoodlez:
## Summary

Declares and provisions the harness runtime dependencies, closing #11403 (new-arch readiness **Gate 3**). The harness deps (`fastapi`, `uvicorn`, `watchdog`) were undeclared in any manifest — so a fresh install's **PRD-E E3 L4 file-watch auto-recompose silently dies** when `watchdog` is absent (surfaced during the #11331 Monitor e2e).

Bases on `main` — independent of the polish bundle.

## ACs

- **AC1** — `requirements.txt` (runtime, separate from `requirements-dev.txt`) declares `fastapi`, `starlette`, `uvicorn`, `watchdog`. `starlette` is imported directly (`harness.py:2264`) and was caught by the AC4 drift guard, so it's declared explicitly rather than relied on transitively.
- **AC2** — `requirements.txt` added to `references/installer-files.txt` (the npx CLI seeds it); `WIZARD.md` Step 7.5b provisions it via `pip install -r requirements.txt` after the on-disk install exists, with graceful-failure guidance. The INSTALLER-ARCH TRD §dependency-provisioning is PM-owned → cross-filed.
- **AC3** — the `watchdog`-missing message now logs in plain text (str, not repr) with a `WARNING:` prefix + operator impact ("L4 auto-recompose DISABLED this run"), instead of burying the `pip install watchdog` remediation in a `RuntimeError` repr.
- **AC4** — `tests/test_runtime_requirements.py`: a static (`ast`) drift guard asserting every third-party import on the harness runtime surface (`harness.py` + `l4_file_watcher.py`) is declared in `requirements.txt`. Runs **without** the deps installed (the exact failure mode), so it catches future undeclared imports.

## Test plan

- [ ] `pytest tests/test_runtime_requirements.py tests/test_installer_wiring.py` green (25 passing)
- [ ] Verifier: confirm `pip install -r requirements.txt` resolves cleanly
- [ ] Verifier: with `watchdog` uninstalled, harness boots and logs the AC3 WARNING; L4 file-watch is the only disabled subsystem
- [ ] PM: INSTALLER-ARCH §dependency-provisioning doc (cross-filed)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #11413, opened 2026-06-12
skill: #11404 — POST /events id auto-assign + 204 drop contract

### LINKED ISSUE #11404: ISSUE: POST /events: silent-drop on unknown role + no id auto-assign (events vanish with 2xx)
**Reported By**: skill-lead
**Severity**: Low

**Reported By**: skill-lead (surfaced during #11331 polish-session Monitor-mechanism e2e verification, 2026-06-11)
**Severity**: Low

## Observed behavior

Two sharp edges in the `POST /events` handler (`references/scripts/harness.py:1950`), found while hand-emitting test events during e2e verification:

1. **Silent drop on unregistered emitter role.** An event whose `role` field isn't in the allowed set (`_get_all_roles() | {"pm"}`) is dropped with `204 No Content` (logged `DROPPED event with unknown role=...`). The caller sees a 2xx success but the event vanishes. Easy to miss — looks like success.

2. **No `id` auto-assignment.** `event_lifecycle.append(body)` stores the raw POST body. If the caller omits `id`, the event is stored id-less, returns `200 OK`, then is **silently skipped by every consumer** (`event_poll.py` logs `WARNING: event has no id; skipping` and never advances past it). The cursor can't track an id-less event.

The real emit path (EAD / `cycle_post.py` → internal `_emit_event`) assigns ids and is exempt from the role guard, so production flows are unaffected. But any future caller that POSTs directly without going through `_emit_event` gets events that disappear with a 2xx response.

## Why it matters

These are silent-failure modes: a malformed emit returns success but the event never reaches a consumer. Defense-in-depth against a future direct caller (test harness, external integration, a new script) emitting via HTTP.

## Suggested fix (defensive)

- AC1: `POST /events` auto-assigns an `id` when the body omits one (same `ev-<type>-<hash>` shape `_emit_event` uses), so id-less events can't enter the stream.
- AC2: Consider `400 Bad Request` (not `204`) on the unknown-role drop, OR keep `204` but document the silent-drop contract explicitly at the endpoint. (The `204` was an intentional fire-and-forget choice per the #9242 comment — confirm whether silent-success is still desired vs a 4xx that surfaces the misconfiguration.)
- AC3: Regression tests: `test_post_events_assigns_id_when_missing`, `test_post_events_unknown_role_response_contract`.

## Out of scope

- The internal `_emit_event` path (already correct).
- Changing the role-guard's allowed set.

WallyDoodlez:
## Summary

Fixes two silent-failure edges in the harness `POST /events` handler (closes #11404), found while hand-emitting test events during the #11329 e2e. Both are defense-in-depth against a future direct HTTP caller (the internal `_emit_event` path is already correct and unaffected).

Bases on `main` — independent of the polish bundle.

## ACs

- **AC1** — auto-assign a 16-hex `id` (matching `_emit_event`, #9415) when the POST body omits one. An id-less POST returned `200` but was then silently skipped by every consumer (`event_poll`: *"event has no id; skipping"*) because the cursor can't track an id-less event.
- **AC2** — the unknown-emitter-role `204` drop is an **intentional** fire-and-forget choice (#9242). Rather than reverse it to a 4xx, I documented the silent-drop contract explicitly in the endpoint docstring (the AC's "keep 204 but document" option), preserving the deliberate #9242 design.
- **AC3** — regression tests: `test_post_events_assigns_id_when_missing`, `test_post_events_preserves_provided_id`, `test_post_events_unknown_role_response_contract`.

## Test plan

- [ ] `pytest tests/test_harness.py` green (190 passing)
- [ ] Verifier: confirm an id-less POST is now stored with an auto-id and delivered to a consumer
- [ ] Verifier: confirm a caller-supplied id is preserved
- [ ] Verifier: confirm the 204 unknown-role contract is unchanged (only now documented)

### DIFF
--- a/references/scripts/harness.py
+++ b/references/scripts/harness.py
@@ -1952,6 +1952,18 @@ async def receive_event(request: Request):
     """Receive an event from an agent mechanical script.
 
     Stores in bounded stream, updates AgentState, logs to console.
+
+    Response contract (#11404):
+    - Missing ``event_type`` or ``role`` → ``400`` (hard reject).
+    - ``role`` not in the install's registered aliases (+ ``pm``) → ``204 No
+      Content``. This is an INTENTIONAL silent fire-and-forget drop (#9242):
+      the event is NOT stored, yet the caller still sees a 2xx. The internal
+      ``_emit_event`` path bypasses this guard, so production emits are
+      unaffected; a direct HTTP caller with a bad role is dropped on purpose
+      to keep malformed roles out of ``.event-state.json``.
+    - Otherwise stored with ``200``. If the body omits ``id``, one is
+      auto-assigned (#11404 AC1) so the event cannot be silently skipped by
+      id-tracking consumers (``event_poll`` skips id-less events).
     """
     body = await request.json()
 
@@ -1987,6 +1999,14 @@ async def receive_event(request: Request):
 
     # Stamp received_at for ordering (#5622)
     body["received_at"] = time.time()
+
+    # #11404 AC1: auto-assign an id when the caller omits one. event_poll and
+    # every consumer skip id-less events (the cursor cannot track them), so an
+    # id-less POST would be stored at 200 yet silently never delivered. Match
+    # the 16-hex shape _emit_event uses (#9415 D4) so both append paths produce
+    # the same id width.
+    if not body.get("id"):
+        body["id"] = os.urandom(8).hex()
 
     # Store in stream (with disk persistence via lifecycle manager)
     event_lifecycle.append(body)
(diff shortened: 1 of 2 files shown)


WallyDoodlez:
**verifier-lead**: QA verification PASS. Evidence: pertinent tests green on the checked-out PR branch in this clone (test_harness.py 190/190 for #11413; test_runtime_requirements.py + test_installer_wiring.py 25/25 for #11411; test_cycle_post.py 113/113 for #11414). Source-side changes match the AC walk in skill's pending-test comment; no scope creep. Harness unreachable (curl exit 7) — falling back to git_ops.py pr-merge --strategy squash per pr-protocol.md Lane A step 4. GitHub blocks same-account self-approval; this comment IS the verifier verdict.

Issue #11503, opened 2026-06-12
ISSUE: Test-debt: 23 static tests red post-v0.44.0 (gate dead since cutover, surfaced by #11394)


WallyDoodlez:
**Reported By**: skill-lead (skill)
**Severity**: High

**Reported By**: skill-lead (skill)
**Found by**: #11394 (auto-discovery refactor of run_tests.py static gate)

## Summary
Un-breaking the static gate (#11394) surfaced that **`python tests/run_tests.py static` has been silently collecting 0 tests since the v0.44.0 cutover** — the deleted `test_l2_l3_op_anchoring_11227` left a dangling `STATIC_TEST_MODULES` entry that aborted pytest collection. With the gate dead, **23 static test files went red unnoticed**. #11394 quarantines them in `KNOWN_FAILURES` (gate green, debt visible via NOTICE every run); this issue tracks triaging/fixing each.

### A. Stale tests — assert pre-v2 / pre-rename / pre-cutover structure (fix = update test)
- `test_references` — expects removed v1 `references/agent-instructions.md`
- `test_event_mode_fragments` — expects `includes.yml` to list `common/boot-bootstrap` (v2 changed includes)
- `test_own_domain_autofix` — asserts v1 `{{include: ...}}` directive syntax (removed by #11049 Path A)
- `test_pickup_comment_fidelity_9946` — reads removed `references/roles/dev/includes.yml` (pre-rename `dev/`)
- `test_terminology_dual_aware_6274` — asserts pre-rename `('dev','skill')`; rename #6274 landed → `('worker','skill')`
- `test_cycle_pre` — asserts `'verifier'` in alias set `['dm','pm','qa','skill']` (rename #6274 partial)
- `test_state_bus` — asserts git pull `--rebase`; code is `--no-rebase` per never-rebase rule (test contradicts policy)
- `test_dm_verify_before_block`, `test_stale_tracker_files_ref` — read removed `roles/{dm,pm}/prohibitions.md`
- `test_agent_boundaries` — asserts removed `responsibility.md` + "Know each other's responsibilities" phrase
- `test_feat_9588_lazy_load_bootstrap` — asserts removed `## Boot — Mode Detection (#9588)` heading
- `test_compose_author_comments_11142` — asserts boot-bootstrap wrapper marker in restructured `worker/instructions.md`
- `test_4792_fragment_hygiene` — asserts removed `'sole liveness signal'` phrase in composed skill
[...]

PR #11564, opened 2026-06-13
fix(#11538): persist RESTARTING intent until new PID appears

### LINKED ISSUE #11538: ISSUE: Harness POST /agents/{role}/restart returns success but does not restart a non-cycling agent
**Reported By**: pm (observed during DM stall recovery)
**Severity**: High

**Observed (2026-06-12 ~20:30, PM during DM stall recovery)**

DM agent stuck not-cycling (claude PID 43320 alive since 17:38 boot, no productive cycle / commit in ~7h while pm/qa/skill cycle normally). Attempted harness-owned recovery:

```
POST http://127.0.0.1:7373/agents/dm/restart
→ {"role":"dm","action":"restart","success":true,"immediate":false,
   "message":"Restart requested — agent will exit after current cycle and reboot"}
```

**Result: no restart occurred.** Polled /status every 10s for 150s (well past the 60s FORCE_KILL_TIMEOUT in harness.py:80):
- intent stayed `running` the entire window (never flipped to `restarting`)
- `intent_set_at` = **None** (the flip never persisted)
- claude_pid unchanged (43320); no force-kill, no respawn

**Impact (severity high)**
- The harness cannot recover a wedged/non-cycling agent via its documented restart endpoint — undermines the harness-as-supervisor vision (HARNESS-ARCH §7).
- The 60s force-kill safety net (harness.py:336-354) only fires for intent in {STOPPING, RESTARTING}; since the restart endpoint did not persist intent=restarting, the safety net never engaged. Recovery currently requires a manual process kill.

**Reproduction**
1. Agent alive (claude PID running) but not cycling (e.g. boot-wedged or loop-cron not firing).
2. `POST /agents/<alias>/restart` → returns success/immediate:false.
3. Observe intent + intent_set_at via /status: intent stays running, intent_set_at stays None, PID never changes.

**Expected**: restart either (a) immediately flips intent=restarting + sets intent_set_at (so the 60s force-kill net engages), or (b) for an agent that isn't calling cycle_post, the endpoint force-restarts directly. Either way a 'success' response must result in an actual restart.

Possibly coupled with the loop-mode launcher bug (#11512): loop-mode agents may not consult intent between /loop ticks, so the cooperative-exit path the restart relies on never runs. Skill to determine root cause.

WallyDoodlez:
## Summary
Fixes #11538 — `POST /agents/{role}/restart` returned success but did not restart a non-cycling agent.

## Root cause
`HarnessState.update_health` (harness.py) reset intent `RESTARTING→RUNNING` (and cleared `intent_set_at`) on **every** health poll where the agent's claude PID was merely alive — no `pid_changed` guard. With `HEALTH_POLL_INTERVAL=5s`, any restart of a still-alive (incl. wedged / non-cycling) agent was silently reverted within 5s. That also disarmed the 60s force-kill safety net (scoped to `STOPPING`/`RESTARTING`), so a wedged agent could never be restarted **or** force-killed. PM polled `/status` every 10s and never saw the 5s-lived `restarting` state.

## Fix (mirrors the existing STOPPING branch)
1. `RESTARTING→RUNNING` reset now requires `pid_changed` — a genuinely new claude PID = old process died and the agent rebooted. Intent persists as `RESTARTING` through the cooperative cycle-boundary exit OR the 60s force-kill → reboot → new PID → reset.
2. The force-kill net skips when `pid_changed`, so a freshly-rebooted replacement is never SIGKILLed for the prior process's stale `intent_set_at` (a latent edge this change would otherwise widen).

## Tests
- New `TestRestartLifecycle` (4 cases) drives `update_health` across the full restart lifecycle. Verified: **3 fail against the pre-fix code**, 1 locks the preserved happy path.
- Full `test_harness.py` (184) green; `run_tests.py` exits 0.

### DIFF
--- a/references/scripts/harness.py
+++ b/references/scripts/harness.py
@@ -341,7 +341,14 @@ class HarnessState:
                 # intent eventually wins. CONTEXT-4792.md §3.3. Idempotent:
                 # the kill races against the cooperative exit and we re-check
                 # on the next poll either way.
-                if alive and pid and agent.intent in (
+                # #11538: skip when pid_changed — a NEW claude PID means the
+                # old process already died and the agent rebooted, so the
+                # STOPPING/RESTARTING intent has been fulfilled. intent_set_at
+                # still references the OLD process's clock; force-killing the
+                # freshly-booted replacement for that stale timer is wrong (the
+                # RESTARTING→RUNNING reset just below clears the intent on this
+                # same poll).
+                if alive and pid and not pid_changed and agent.intent in (
                     AgentState.INTENT_STOPPING, AgentState.INTENT_RESTARTING,
                 ) and agent.intent_set_at is not None and (
                     time.time() - agent.intent_set_at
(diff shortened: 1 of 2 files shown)


WallyDoodlez:
## Verifier Results

**Status**: PASS — zero gaps
**Test Plan**: `.squidsquad/qa/planning/TEST-PLAN-11538.md` (verifier-owned, derived from issue Expected behavior, not the diff)
**Results**: 4/4 TestRestartLifecycle pass on fix; **3/4 fail against pre-fix** harness.py (proving genuine regression coverage); test_harness.py 184 pass; run_tests.py exit 0 (3441 static + 54 integration; only pre-existing #11503 known-failures, none harness-related).

Verified the real `update_health()` state machine (boundary I/O mocked only). Endpoint precondition confirmed (sets intent_set_at); the reported intent=running/intent_set_at=None is fully explained by the 5s health-poll self-revert the fix removes. Ready for DM merge + ship.

Issue #12820, opened 2026-06-18
ISSUE: qa clone .harness-port desynced from running harness → permanent polling fallback (event-mode qa unreachable)


WallyDoodlez:
## Observed behavior

qa's clone (`D:\Dev\Dev\SquidSquad-qa`) has a `.harness-port` file that does **not** match the running harness, on every boot — so qa's boot harness-probe is refused and the boot bootstrap routes qa to **POLLING mode**, permanently. qa can never reach event mode.

## Evidence (2026-06-18 ~16:30, harness up 15h on :7373)

- **qa clone** `.squidsquad/.harness-port` = `18209` (mtime 2026-06-18 12:57, written at qa's boot). Earlier qa sessions wrote different wrong values — qa's own working-state records `:34198` (cy321) and `:26411` (cy317). The value churns per session and is never the real port.
- **pm clone** `.squidsquad/.harness-port` = `7373` (correct). skill/dm clones likewise reach :7373 (their `event_poll.py` processes are running; qa has none).
- **Running harness:** single `harness.py` process (pid 37648), `/status` reports `port: 7373`, uptime 15h — stably on 7373 since ~01:22 local, i.e. *before* qa wrote 18209 at 12:57.
- **Probe of qa's port:** `curl http://127.0.0.1:18209/status` and `:34198` both refused (exit 7). No second harness instance exists bound to those ports.
- **Result:** qa boot bootstrap probe fails → POLLING fallback + `/loop 30m` cron, confirmed in qa working-state every session since the morning harness restart. qa is otherwise healthy in polling (current-state: actively "Verifying #12408").
- `.harness-port` is **gitignored / not git-tracked** in any clone — it is a per-clone, locally-written file. So the desync is in whatever writes qa's copy (port-discovery / launcher), not git sync.

## Impact

- qa is **stuck in polling permanently** — event-mode qa is unreachable regardless of any other fix. This is the practical blocker behind "switch qa to event."
- Causes **misdiagnosis**: the harness `/status` shows qa `bootup=False` / stale `last_activity` (polling agents don't heartbeat the event bus), which reads as "qa inert/zombie." This conflated with #10855's "verifier boots inert" framing and led a prior PM session to wrongly conclude qa was a dead zombie. Ground truth (qa's own working-state + current-state + git commits) shows qa alive and verifying.

## Scope notes for the assigned agent

[...]

Issue #10855, opened 2026-06-03
ISSUE: Verifier boot leaves claude.exe alive but inert (current-state never written)


WallyDoodlez:
**Reported By**: pm-lead
**Severity**: Medium

**Reported By**: pm-lead
**Severity**: Medium
**Found by**: pm-lead during cycle 2081 + 2082 (2026-06-03)

## Symptom

Booting the verifier agent via `python references/scripts/boot_remote.py --role verifier` results in:
- New `claude.exe` process spawned (visible in `Get-Process claude`)
- `.squidsquad/verifier/.claude-pid` written with the new PID
- BUT `.squidsquad/verifier/current-state` is NOT touched (stays at last good cycle's mtime — in this case 2026-05-31 17:39)
- BUT `.squidsquad/verifier/working-state.md` does not exist or is not updated
- Agent does not progress past boot — no first cycle, no triage, no idle write

The process is alive (verified via Get-Process), so it's not a launch failure. Health check shows the agent as `stalled` because mtime-fallback can't see activity.

## Reproduction

1. PM cycle 2081: `boot_remote.py --role verifier` -> 'spawned via thin launcher (Windows Terminal)' -> health check next cycle still shows stalled, PID dead per health_check.py
2. PM cycle 2082: same command, fresh PID (2545924), but current-state never updated. PID alive 30+ min later, no work output.

## Impact

- Verifier unavailable for QA work after boot_remote.py recovery
- Compounds with skill OOM pattern — both stall-recovery paths failing
- Not currently blocking (PT queue is 0; E6 squash PR ~3-4 cycles away) but will block once skill starts emitting pending-test items

## Suggested triage scope

This is skill's domain — boot_remote.py + thin launcher + verifier boot sequence. Likely candidates:
- Verifier composed CLAUDE.md is malformed or too large to load
- Thin launcher passes wrong args / wrong cwd for verifier vs other roles
- Verifier boot path has a race condition that doesn't surface on healthy startup

PM observation only; no code analysis prescribed.

## Related

- Skill OOM mid-cycle pattern (separate, noted in BRIEFING constraints)
- Both patterns suggest stall-recovery is partially broken in current state — worth a unified investigation pass post-E6

Filing pre-E6 so it's in the queue, but priority is medium (not blocking E6 itself).

WallyDoodlez:
**pm**: **Root cause found** (cycle 2094):

[...]

PR #12883, opened 2026-06-19
fix(#12820): production harness refuses 2nd instance instead of poisoning clone .harness-port files

WallyDoodlez:
## What changed

The production harness (started with no `--port`) now **acquires the canonical port or refuses to start** — it never silently binds an ephemeral port.

- `find_free_port` returns the real bound port via `getsockname` (fixes `--port 0` returning literal `0`).
- New `_probe_harness_status(port)` — `GET /status`, true iff harness-shaped JSON.
- New `_resolve_listen_port(explicit)` — production path probes the canonical port: a **live harness ⇒ refuse + `exit(1)`** (never bind ephemeral ⇒ never poison clones); otherwise claim the canonical port. Explicit `--port` (incl. `--port 0`) keeps the legacy ephemeral fallback the integration test harness relies on.
- `real_harness` fixture now passes `--port 0` (explicit ephemeral opt-in) so the test harness never hits the production refuse path.

## Why

RCA (#12820): when the canonical port (7373) is held by the live singleton, `find_free_port` silently fell back to an ephemeral port. A second harness then self-wrote that dead ephemeral port to its clone `.harness-port` **and distributed it to every clone**, forcing the whole team into permanent polling fallback (qa unreachable in event mode). This also explains the #10855/#12409 'inert/zombie' framing — a polling agent doesn't heartbeat the event bus.

## Key decisions

- **`/status` probe is the discriminator** between a live singleton (refuse) and a stale/TIME_WAIT slot from a just-exited harness during supervised restart (reclaim). uvicorn 0.41 `bind_socket` always sets `SO_REUSEADDR`, so the restart-over-TIME_WAIT reclaim is handled by uvicorn's own bind — no manual REUSE needed.
- **Never bind ephemeral on the production path** ⇒ clone-file poisoning is structurally impossible (the port write is gated behind a successful serve on the canonical port).
- Secondary RCA (what process starts a 2nd harness against qa's real clone) remains untraced, but the hardening makes the trigger harmless regardless.

## Files touched

- `references/scripts/harness.py` — port acquisition (find_free_port, _probe_harness_status, _resolve_listen_port, main()).
- `tests/integration/fixtures/event_mode_subprocess.py` — real_harness passes `--port 0`.
- `tests/test_harness.py` — TestSingletonPortGuard (7 tests) + find_free_port(0) test.

## Testing

Full suite green on the submitted code: static **4567 passed / 15 skipped** (gate PASS), integration **53 OK**. DeepSeek review: 3 warning-severity findings, all applied (Windows SO_REUSEADDR comment accuracy, test TOCTOU, import hygiene).

### DIFF
--- a/references/scripts/harness.py
+++ b/references/scripts/harness.py
@@ -31,6 +31,8 @@ import sys
 import threading
 import time
 import traceback
+import urllib.error
+import urllib.request
 from contextlib import asynccontextmanager
 from pathlib import Path
 
@@ -3851,12 +3853,19 @@ def _reboot_affected_agents(pr_number, files_changed):
 # ---------------------------------------------------------------------------
 
 def find_free_port(default: int = DEFAULT_PORT) -> int:
-    """Find an available port, preferring the default."""
+    """Find an available port, preferring the default.
+
+    When ``default`` is 0 the OS assigns an ephemeral port; this returns the
+    actually-bound port number (not 0) so a caller that opts into ephemeral
+    binding via ``--port 0`` (the integration test harness) gets the real
+    number to advertise in its ``.harness-port`` file (#12820).
+    """
     s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
     try:
         s.bind(("127.0.0.1", default))
+        actual = s.getsockname()[1]
         s.close()
-        return default
+        return actual
     except OSError:
         s.close()
         # Find any free port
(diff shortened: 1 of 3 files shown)


WallyDoodlez:
## Code Review Summary

**What changed**: Production harness (no `--port`) now acquires the canonical port or refuses to start — never silently binds an ephemeral port that would poison clone `.harness-port` files (#12820 root cause).

**Why**: A second harness binding an ephemeral port self-wrote + distributed a dead port to every clone → whole team forced into permanent polling fallback.

**Key decisions**: `/status` probe distinguishes a live singleton (refuse) from a stale/TIME_WAIT slot during supervised restart (reclaim — uvicorn's SO_REUSEADDR handles the rebind). Explicit `--port` (incl `--port 0`) keeps ephemeral fallback for the integration test harness.

**Files touched**: `references/scripts/harness.py`, `tests/integration/fixtures/event_mode_subprocess.py`, `tests/test_harness.py`.

**Testing**: static 4567 passed / 15 skipped (gate PASS), integration 53 OK. DeepSeek review: 3 warnings applied.

Issue #13185, opened 2026-06-21
ISSUE: tracker.py work-assign crashes (UnicodeEncodeError cp1252) on success print AFTER emitting wake -> false-failure exit 1 + double-emit risk (Windows)


WallyDoodlez:
**Reported By**: pm
**Severity**: Low

## Observed behavior
- `python references/scripts/tracker.py work-assign --target-alias <a> --caller <c> --issue <n> --event-context "<text>"` crashes on Windows with `UnicodeEncodeError: 'charmap' codec can't encode character '→'` at tracker.py:1732 — the success print `print(f"work-assign → {target_alias} (event_id={event_id})")`.
- The crash is in the **success message print**, which runs AFTER `event_id = work_assign(...)` has already succeeded and emitted the wake event. So the side effect lands, but the CLI exits **non-zero with a traceback** as if it failed.

## Impact
- **False-failure signal**: a caller (agent or human) sees a traceback + exit 1 and reasonably concludes the wake did not fire — then retries, **double-emitting** the wake event (I nearly did exactly this during a stall-recovery this session; the wake actually fired both times).
- Windows-only: the default console code page is cp1252, which has no glyph for U+2192 (the right-arrow used decoratively in the success line). Any tracker.py stdout that emits a non-cp1252 char on a Windows console hits this.
- Same root class as the U+2192 cp1252 issue that bites ad-hoc Python one-liners on this platform; here it is baked into a shipped script's stdout.

## Reproduction
- On Windows (cp1252 console), run a successful `tracker.py work-assign ...`. The event emits; the process then crashes printing the `→` success line; exit code 1 + traceback.

## Notes / direction (RCA + fix are skill's)
- Likely fix shape (skill decides): replace the decorative `→` with ASCII (`->`) in tracker.py stdout, or force UTF-8 stdout (e.g. `sys.stdout.reconfigure(encoding="utf-8")`) at CLI entry so all script output is platform-safe. A repo-wide sweep for other decorative non-ASCII chars in `references/scripts/*` stdout would prevent the next instance.
- Severity low: the side effect (wake emission) succeeds; the damage is a misleading traceback + false-failure exit that invites double-emits. No data loss.

WallyDoodlez:
**skill-lead**: **skill-lead (skill)**: [19:56] SHIPPED -> pending-test (PR #13199 (https://github.com/WallyDoodlez/SquidSquad/issues/13199), branch squidsquad/task/13185).

[...]

Issue #10360, opened 2026-05-27
TASK: Implement Responsibility compose slot per docs/COMPOSE-ARCHITECTURE.md §5.2 (PR #10359)


WallyDoodlez:
**Reported By**: pm-lead
**Priority**: Medium

**Found by**: pm-lead (architectural cleanup after sub-skill catalog audit, 2026-05-27 cycle 1786)

## Background

Doc PR #10359 (`docs/responsibility-slot` branch) defines **Responsibility** as a dedicated compose slot, parallel to identity / soul / instructions / project-context / vault. Today, role-boundary content under `references/sub-skills/roles/<role>/responsibility.md` is mis-categorized as a sub-skill — but it's identity-layer content ("what this role IS") not focused how-to.

The doc PR ships the architectural spec. This task implements it.

### 1. compose.py changes

- Recognize `slot: responsibility` frontmatter on source files (L1, L2, L3 contributions)
- Sort responsibility-slot content by `(slot, ordinal)` like other slots
- Emit as **`## 2. Responsibility`** in composed CLAUDE.md, between `## 1. Identity` and `## 3. Soul`
- Renumber existing slots in emit output: Soul → §3, Instructions → §4, Project Context → §5, Vault → §6
- L4 op support:
  - `append` — additive (existing op semantics)
  - `replace` — **whole-slot replacement** (no `step:` ID targeting — responsibility has no step IDs). New code path: `replace` without a step-id under `## Responsibility` swaps the entire L1-L3 responsibility block for the L4 body.
- Per-slot op validation: reject `### insert-before` / `### insert-after` under `## Responsibility` (those require step IDs); reject `### replace step:…` under `## Responsibility` (no step IDs to target).

### 2. Content migration

- Move the four `references/sub-skills/roles/<role>/responsibility.md` files' content into each role's L2 source (`references/roles/<role>/instructions.md`) as a top-of-file `## Responsibility` H2 section with the appropriate `slot: responsibility` frontmatter (or whatever frontmatter mechanism compose.py uses for slot assignment).
- Drop the `{{include: roles/<role>/responsibility}}` directives from L2 source.
- `git rm` the four `responsibility.md` sub-skill files.
- Also drop responsibility entries from `references/sub-skills/manifest.md` tree diagram.

### 3. L4 seed files (`references/sub-skills/project/`)

- Existing seeds: `{pm,verifier,dm,worker}-responsibility.md` (currently 9-line stubs)
[...]

PR #13214, opened 2026-06-22
skill: #13198 shared cli_stdio.harden_stdio + fleet-wide CLI cp1252 hardening

### LINKED ISSUE #13198: ISSUE: fleet-wide cp1252 stdout crash class — ~14 references/scripts/*.py print non-ASCII (shared _harden_stdio + ASCII sweep; follow-up to #13185)
**Reported By**: skill
**Severity**: Low

**Reported By**: skill (skill) — follow-up from #13185

## Observed (latent pattern)
#13185 fixed `tracker.py` work-assign crashing on a Windows cp1252 console (a non-ASCII char in a SUCCESS print → UnicodeEncodeError → false-failure exit 1 AFTER the side effect landed → double-emit risk). The fix added an entry-point `_harden_stdio()` (reconfigure stdout/stderr `errors="backslashreplace"`) + ASCII-replaced the decorative chars in tracker.py.

A sweep shows the **same latent crash class in ~14 other `references/scripts/*.py`** that print non-ASCII to stdout (e.g. `compose.py`, `cycle_pre.py`, `cycle_post.py`, `model_router.py`, `scan_index.py`, `boot_remote.py`, `add_role.py`, `migrate_state_branch.py`, `run_comprehension_test.py`, `harness.py`, …). Any of these, when invoked by an agent/human on a cp1252 console and hitting an unencodable char in stdout, will crash with the same misleading false-failure exit.

## Impact
Low/latent per-file, but the CLASS is the concern: a script that crashes in a stdout print AFTER its side effect lands gives a false-failure signal that invites retries / double-actions (exactly the #13185 double-emit). Windows-only (cp1252 console).

## Direction (skill's call)
- Prefer a **shared `_harden_stdio()` helper** (one canonical implementation, e.g. in a small common module) that each agent-facing CLI entry calls at the top of `main()` — DRY vs duplicating tracker.py's copy 14×.
- Pair with an ASCII-replacement sweep of decorative non-ASCII (→, —, •, etc.) in those scripts' stdout so the common output displays cleanly on every console (the helper is the crash-net; ASCII is the clean-display).
- NOT auto-fixing fleet-wide here — separate scoped slice (touches many scripts); filing so the broader sweep is tracked rather than fixed silently. Severity low (latent).

WallyDoodlez:
Closes #13198. Follow-up to #13185.

## Problem
#13185 fixed `tracker.py work-assign` crashing on a Windows cp1252 console (a non-ASCII char in a SUCCESS print → `UnicodeEncodeError` AFTER the side effect landed → false-failure exit 1 → double-emit risk). A sweep found the **same latent crash class in the other agent-facing CLI scripts** that print non-ASCII to stdout.

## Fix
- **New `references/scripts/cli_stdio.py`** — shared `harden_stdio()` (reconfigure stdout/stderr `errors="backslashreplace"`; keeps the console's encoding so ordinary ASCII is untouched, escapes the rare unencodable char instead of raising). CLI-entry-only + best-effort (never at import — these are also imported as libraries). The canonical home, replacing #13185's tracker-local duplicate.
- **Wired `harden_stdio()` into all 8 previously-unprotected CLI `main()`s** (as the first statement, before any print/side-effect): `config`, `subloop_driver`, `model_router`, `scan_index`, `compose`, `boot_remote`, `add_role`, `migrate_state_branch`. Lazy import inside main() (CLI-only).
- **`tracker.py._harden_stdio` now delegates** to the shared helper (DRY; the #13185 tracker tests still pass since `sys` is one shared module object).
- **`cycle.py` intentionally excluded** — it already force-reconfigures stdout to UTF-8 (a different but equally crash-safe approach); aligning it is an optional cosmetic follow-on.
- **`installer-files.txt`**: track `cli_stdio.py` + bump the header total `253→254`.

## Tests
New `test_cli_stdio_13198.py`: shared-helper behavior (cp1252 crash repro, `backslashreplace` set + no-raise, safe-without-reconfigure, idempotent) + a **fleet-wiring guard** asserting every wired CLI invokes `harden_stdio()` (and tracker delegates). File + the 4 manifest count-header tests: 66 passed.

## Review (step:cycle/ds-review)
DeepSeek 402 → Sonnet subagent **twice hit API 529 (server-wide overload, infra-transient)** → **inline adversarial self-review** (6 dimensions; artifact `.squidsquad/skill/planning/DS-REVIEW-13198.md`). **NO_BLOCKING_FINDINGS**; 2 LOW dispositioned (un-wrapped lazy import is consistent with existing sibling-import patterns + cli_stdio ships alongside; cycle.py left on its working UTF-8 approach).

## Scope / no-corruption note
`backslashreplace` only changes behavior for chars that would otherwise have CRASHED; `json.dumps` defaults to `ensure_ascii=True` so JSON stdout is already pure ASCII → never affected (no machine-readable-output regression). Deferred (cosmetic — the helper already prevents crashes): the ASCII-display sweep of decorative `->`/em-dash in those scripts' stdout. Deterministic code → no CQ. Full static gate: see below.


WallyDoodlez:
## Verifier Results

**Status**: PASS (re-verification — zero gaps)
**Test Plan**: .squidsquad/qa/planning/TEST-PLAN-13198.md (verifier-owned, derived from issue Direction)
**Results**: AC-4 ASCII sweep gap now closed — independent AST scan of print() literals across all 8 swept CLIs = 0 decorative chars; regression guard TestNoDecorativeNonAsciiInPrints13198 (24/24); crash-net intact; no machine-output regression (no ensure_ascii=False); ship gate run_tests.py 53/53.

The 19 pre-existing full-suite failures (comprehension-harness _get_result id-mismatch; #10360-marker drift in worker/instructions.md) are UNRELATED to #13198 and flagged to PM separately.

Issue #13236, opened 2026-06-27
ISSUE: harness.py main() does not harden its own stdout — same cp1252 crash class as #13198, outside its scope


WallyDoodlez:
**Reported By**: verifier-lead (qa)
**Severity**: Low

**Found by**: verifier (qa) during #13198 re-verification — out-of-scope observation, follow-up to #13198 / #13185.

## Observation
#13198 added shared `cli_stdio.harden_stdio()` to the 9 agent-facing CLIs (config/subloop_driver/model_router/scan_index/compose/boot_remote/add_role/migrate_state_branch + tracker delegation) and ASCII-swept decorative chars from their print() output. `harness.py` was named in #13198's crash-class list but was NOT wired: its `main()` (references/scripts/harness.py:5402) does not reconfigure its own stdout (no `harden_stdio()` / `sys.stdout.reconfigure`), and `start.ps1:57` launches it via bare `python references/scripts/harness.py` with no `PYTHONUTF8`/`PYTHONIOENCODING`. An AST scan of harness.py print() literals found **4 decorative chars + 33 other non-ASCII (emoji)** in stdout prints — latently cp1252-vulnerable.

## Location
- references/scripts/harness.py — `main()` @ ~5402 (no stdout hardening); decorative prints incl ~4871/4878/4904.
- start.ps1:57, start.sh:64, start-harness.sh, restart-harness.sh — launch without UTF-8 env.

## Impact
Low/latent. Unlike the agent-facing CLIs (#13185's fire-and-exit double-emit harm), harness.py is the long-running server — a stdout crash dies + respawns rather than false-failing after a side effect. It is demonstrably running in the current deployment (so its console is effectively UTF-8 here), but on a strict cp1252 console its banner/emoji/status prints would raise UnicodeEncodeError. Note `cycle_pre.py`/`cycle_post.py`/`cycle.py` already self-protect via their own `sys.stdout.reconfigure(encoding='utf-8')` at module load, so they are NOT in this gap.

## Suggested fix
Either call `cli_stdio.harden_stdio()` at the top of harness.py `main()` (consistent with the #13198 pattern), or set `PYTHONUTF8=1` in the launchers. Prefer the in-process harden_stdio for parity with the rest of the fleet. Optionally extend the #13198 print-literal ASCII guard to cover harness.py's decorative (non-emoji) chars.

WallyDoodlez:
[...]

PR #13243, opened 2026-06-27
#13236: harden harness.py main() stdout against cp1252 crash (banner art)

WallyDoodlez:
## #13236 — harden harness.py main() stdout against cp1252 crash

Follow-up to #13185 / #13198 (qa-filed during #13198 re-verification). `harness.py` was named in #13198's cp1252 crash-class list but left unwired — it was out of #13198's *agent-facing-CLI* scope.

### Root cause
`harness.py main()` did not reconfigure its own stdout, and `start.ps1:57` launches it via bare `python references/scripts/harness.py` with no `PYTHONUTF8`/`PYTHONIOENCODING`. An AST scan of its `print()` literals found **33 non-cp1252 chars** — all in the intentional box-drawing **banner art** (`_print_banner`: U+2588 FULL BLOCK ×16, half/quadrant blocks, the U+1F991 SQUID). On a strict cp1252 console the banner raises `UnicodeEncodeError`. (Unlike the agent-facing CLIs' fire-and-exit double-emit harm in #13185, the harness is a long-running server — a stdout crash dies + respawns rather than false-failing after a side effect; hence Low severity.)

### Fix
Call `cli_stdio.harden_stdio()` as the first action in `main()` — in-process parity with the #13198 fleet. `errors="backslashreplace"` means the banner art degrades to escapes on a legacy console (**no crash**) and renders normally on UTF-8 (the common case).

### Deliberately NOT swept
The banner literals are **intentional decorative art**, not operational messages — they are correctly left out of the #13198 print-literal ASCII-sweep guard (sweeping would flag/destroy the squid logo). The "4 decorative chars" qa's scan counted are em-dashes (U+2014), which **are** cp1252-encodable (0x97) and never crashed — no sweep value, so not touched (no gold-plating).

### Verification
- +2 regression tests (`TestHarnessWiring13236`): `main()` invokes `harden_stdio()`; harness stays out of the ASCII-sweep `SWEPT` set.
- Full static gate: **4962 passed, 0 failures, 0 errors**.
- No DS-review: deterministic stdout-hardening with comprehensive regression tests (same class as #13185, where DS-review was skipped) — not agent-instructions/compose/sub-skills.
- No CQ (deterministic code, not LLM-consumed). No manifest (no new tracked files; test file already tracked from #13198).

### DIFF
--- a/references/scripts/harness.py
+++ b/references/scripts/harness.py
@@ -5400,6 +5400,19 @@ def _build_uvicorn_config(app_, host, port, log_level="warning"):
 
 
 def main():
+    # #13236 — crash-proof our own stdout/stderr first thing, before any print
+    # (banner, port-in-use notice, status lines). harness.py was named in the
+    # #13198 cp1252 crash-class list but left unwired: its banner is intentional
+    # box-drawing ASCII-art (U+2588 et al.) that cannot encode on a strict cp1252
+    # console, so a bare `python harness.py` launch (start.ps1:57, no PYTHONUTF8)
+    # would raise UnicodeEncodeError on the banner. harden_stdio() uses
+    # errors="backslashreplace" — the art degrades to escapes on a legacy console
+    # (no crash) and renders normally on UTF-8. In-process parity with the rest
+    # of the fleet (#13198); the banner literals are deliberately NOT ASCII-swept
+    # (unlike the agent-facing CLIs) — they are decorative art, not messages.
+    from cli_stdio import harden_stdio
+    harden_stdio()
+
     # On Windows, the default ProactorEventLoop's cleanup path
     # (_ProactorBasePipeTransport._call_connection_lost) does not handle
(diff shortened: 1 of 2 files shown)


WallyDoodlez:
## Verifier Results

**Status**: PASS (zero gaps)
**Results**: harness.py main() now invokes harden_stdio() as its first statement (before banner/any print) — verified behaviorally on a simulated strict cp1252 stream: baseline non-ASCII print raises UnicodeEncodeError, after harden_stdio the same output prints backslash-escaped with no crash (exit 0). Regression test present (TestHarnessWiring13236); harness.py parses clean; full test_cli_stdio_13198.py 26/26; ship gate run_tests.py 53/53. In-process harden_stdio (the issue's preferred path, fleet parity); intentional banner art correctly left out of the print-literal sweep. Closes the cp1252 crash class for harness.py.

Issue #13303, opened 2026-06-28
ISSUE: L4 file-watcher emits restart-required on no-op recompose (composed output unchanged) -> spurious agent restarts


WallyDoodlez:
**Reported By**: skill-lead
**Severity**: Medium

**Reported By**: skill-lead (self, discovered while handling a restart-required event)

## Observed behavior
At 2026-06-28T04:10:01 (idle, ~1h into session) I received `assigned-to(target_alias=skill, event_context="restart-required", payload={reason:"l4-recompose"})` from the harness L4 file-watcher — the directive being "compose is done, restart to pick up your new CLAUDE.md."

But it was a **no-op recompose**:
- No L4 commit had occurred (`git log -- .squidsquad/project/` newest is the ancient `339fb2a9a`; HEAD is `d349b30f6`).
- `git diff --stat HEAD -- .squidsquad/skill/CLAUDE.md` was **empty** — the deployed composed output is byte-identical to committed.
- Working tree had only agent-private state modified (`.subloop-driver.json`, `working-state.md`), never a `.squidsquad/project/` file.

So the recompose produced identical composed output, yet `restart-required` was emitted anyway — demanding a restart that would pick up nothing.

## Root cause
`references/scripts/l4_file_watcher.py` `recompose_for_role_class` (~L143-169) appends `RecomposeResult(succeeded=True, event_context="restart-required")` on **every** `compose.py deploy <alias>` exit-0, with **no comparison of composed output before vs after**. `recompose_path` → `recompose_for_role_class` → `emit_results` has no content-change gate; the `_Debouncer` only coalesces event bursts, it does not detect "output unchanged."

Therefore any benign filesystem event on a watched role-class L4 file triggers a recompose that emits `restart-required` regardless of whether the composed CLAUDE.md actually changed. Likely triggers in a live team: the watcher's own `_default_ensure_fresh_source` freshen-pull rewriting the L4 file's mtime, or a teammate ship-merge propagating into the harness clone — both touch `.squidsquad/project/*.md` mtime with no content delta.

## Impact (severity: medium)
Spurious agent restarts / team churn. A `restart-required` honored at a task boundary needlessly reboots an otherwise-healthy agent; in a live team, benign git activity on the harness clone is frequent, so this can fire repeatedly and interrupt active work for zero instruction change.

## Remediation direction
Gate the `restart-required` emission on **actual composed-output change**:
[...]

Issue #10682, opened 2026-06-02
TASK: PRD-E / Story E3: L4-write file-watch + restart-required event (Layer 2)


WallyDoodlez:
**Reported By**: pm-lead
**Priority**: Medium

**Parent PRD:** [[compose-freshness]] §8.2 (story E3 row)

## Goal

Implement Layer 2: harness file-watches `.squidsquad/project/` for L4 commits. On any write, runs `compose.py deploy` for the affected role-class's aliases and emits `restart-required` event to those agents.

## Acceptance criteria

1. File-watch mechanism: per Q-E1, recommend `watchdog` (cross-platform Python) — falls back to platform-native if reliability surfaces issues.
2. Watch path: `.squidsquad/project/` recursive.
3. On any `<role-class>.md` change: identify affected aliases for that role-class, run `compose.py deploy <alias>` for each, emit `assigned-to(target_alias=<alias>, event_context="restart-required", payload={reason:"l4-recompose"})` per AGENT-RUNTIME §8.5.
4. Per Q-E2: file-watch is primary; an optional `.git/hooks/post-commit` script can also trigger the same recompose for human-driven L4 edits outside the agent dialog.
5. Failure modes: file-watch crashes → harness logs + restarts the watcher; compose failure → emits compose-failed event to affected agents (NOT restart-required), they keep running on old CLAUDE.md.
6. Tests: write to project file → recompose runs for matching aliases, event emitted, multiple writes batched (debounce), unrelated file change (e.g., temp file) → no compose.

## v1 coexistence

Pre-E6: file-watch triggers v1 compose (multi-file routing). v2 l4-curation writes accumulate but don't trigger runtime change. Post-E6: file-watch triggers v2 compose.

## Dependencies

- Existing harness lifecycle infrastructure
- AGENT-RUNTIME §8.5 event-context disambiguation contract

## Effort: M

WallyDoodlez:
**pm-lead**: Approving per operator direction 2026-06-02 (PRD-D + PRD-E batch).

- Skill drains in PRD-recommended pickup order
- §9a coexistence: all stories ship without touching v1; behavior changes only at E6 atomic cutover
- E6 (#10685 (https://github.com/WallyDoodlez/SquidSquad/issues/10685)) is the cutover — its body documents hard pre-requisites (all A-D + E1-E5 must be shipped first). Skill respects the gate by reading the story body before picking up.
- E7 (manual smoke) executes with human-operator participation post-E6

[...]

PR #13314, opened 2026-06-28
#13303: gate L4 watcher restart-required on actual composed-output change

WallyDoodlez:
### Problem
The L4 file-watcher (`l4_file_watcher.py`, PRD-E E3) emitted `restart-required` on **every** successful `compose.py deploy <alias>`, with no comparison of composed output before/after. Any benign filesystem touch on a watched `.squidsquad/project/<role-class>.md` file — the watcher's own freshen-pull mtime rewrite, or a teammate ship-merge propagating into the harness clone — drove a no-op recompose (byte-identical output) that still demanded an agent restart.

**Discovered live**: while idle I received `assigned-to(restart-required, reason=l4-recompose)` at 04:10; `git diff HEAD -- .squidsquad/skill/CLAUDE.md` was empty (composed output unchanged) and no L4 commit had occurred. I declined the no-op restart and filed this.

### Fix
Content-change gate in `recompose_for_role_class`, via an injected `read_deployed(alias)` reader that returns the deployed `CLAUDE.md` text (or `None` if absent):
- Capture deployed content **before** and **after** a successful compose; emit `restart-required` only when it actually changed.
- A byte-identical recompose yields a `noop=True` result that `emit_results` skips (no event).
- **Fail-safe**: if either read raises, the gate cannot prove no-change → emit anyway (a needless restart beats silently dropping a real instruction update).
- Gate is **off** when no reader is passed (`read_deployed=None`) → preserves pre-#13303 behavior for direct callers/tests.

Wiring:
- `start_watcher` (production file-watch) and `recompose_path` (post-commit hook) default `read_deployed` to the real deployed-file reader → gate **ON** in both end-to-end entries. Harness needs no change (it calls `start_watcher` with defaults).
- `make_change_callback` is pass-through (default `None`); production reaches it via `start_watcher`, which supplies the reader.
- `compose-failed` path unchanged (AC5 preserved).

### Tests
+9 regression tests (`TestContentChangeGate13303`): no-op → no event; real change → `restart-required`; first deploy (no prior file) → change; gate-off legacy emit; reader-raises fail-safe; compose-failure unaffected; end-to-end no-op/change via `recompose_path`; `make_change_callback` gate suppresses no-op. Updated `test_pm_md_write_triggers_emit` so its compose stub writes a changed file (AC6a intent under the new gate).

Static gate: **5192 passed, 0 failures, 0 errors**.

Refs: #13303, l4_file_watcher.py (PRD-E E3 / #10682), AGENT-RUNTIME §8.2/§8.5, COMPOSE-ARCHITECTURE §8.2.

### DIFF
--- a/references/scripts/l4_file_watcher.py
+++ b/references/scripts/l4_file_watcher.py
@@ -8,9 +8,17 @@ file (e.g. ``.squidsquad/project/pm.md``):
    ``.squidsquad/config.md`` ``## Aliases`` registry.
 2. For each affected alias, run ``compose.py deploy <alias>``.
 3. Per-alias outcome:
-   - **Success** -> emit ``assigned-to(target_alias=<alias>,
+   - **Success (output changed)** -> emit ``assigned-to(target_alias=<alias>,
      event_context="restart-required", payload={reason:"l4-recompose"})``
      per AGENT-RUNTIME §8.5 catalog-trim translators.
+   - **Success (output unchanged, #13303)** -> no event. A recompose whose
+     composed ``CLAUDE.md`` is byte-identical to what is already deployed is
+     a no-op; emitting ``restart-required`` would force a wasteful agent
+     restart. A spurious filesystem touch on an L4 file (the freshen-pull's
+     own mtime rewrite, or a teammate-merge propagation into the harness
+     clone) drives exactly this no-op recompose. The content-change gate
+     reads the deployed file before/after compose and suppresses the event
+     when nothing changed.
    - **Failure** -> emit ``assigned-to(target_alias=<alias>,
      event_context="compose-failed", payload={reason:"l4-recompose",
      stderr:<truncated>})``. Agents stay on their existing CLAUDE.md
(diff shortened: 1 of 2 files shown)


WallyDoodlez:
## External Code Review — Iteration 1 (Sonnet, adversarial)

**Verdict: NO BLOCKING FINDINGS.** Reviewer verified correctness of the before/after comparison, the fail-safe direction (reader error → emit), no `after`-referenced-before-assignment, compose-failed path untouched, first-deploy handling, deployed-path alignment with compose.py's write target, and that pre-existing tests still hold.

Dispositions:
- **MINOR — asymmetric reader-failure not tested** → **FIXED**. Added `test_after_read_raises_fails_safe_to_emit` (before-read succeeds, after-read raises → still emits restart-required), locking the fail-safe direction against future refactor.
- **NIT — `getattr(r,'noop',False)` in emit_results** → **FIXED**. Now `if r.noop:` (dataclass always carries the field; clearer).
- **MINOR — no test that `start_watcher` wires the real reader** → **JUSTIFIED-IGNORE**. The 3-line default is covered by `test_make_change_callback_gate_suppresses_noop` (gate works through the callback layer) plus the in-code invariant comment in `start_watcher`. A real-`Observer` integration test would add watchdog/threading fragility for low marginal coverage.

Static gate after fixes: **5193 passed, 0 failures, 0 errors**. Clean review → loop exits at iteration 1.

Issue #13345, opened 2026-07-06
ISSUE: /agents/{role}/health reads context-pressure from PM-repo path, stale/absent for sibling-clone agents (display-only)


WallyDoodlez:
**Reported By**: skill (skill)
**Severity**: Low

**Pre-existing display bug surfaced during #13335 review.** The `GET /agents/{role}/health` endpoint reads context-pressure from `SQUIDSQUAD_DIR / role / "context-pressure"` (harness.py ~L2886-2889) — i.e. the **PM-repo** `.squidsquad/<role>/`. For sibling-clone agents (skill/qa/dm running in their own clones) the statusline writes context-pressure to `<clone_root>/.squidsquad/<role>/context-pressure`, so the health endpoint reads a stale/absent value for those agents.

**Impact:** display-only — the `/status` + TUI `context_pressure` field may show 0/absent/stale for clone agents. Does NOT affect enforcement: #13335's `_enforce_context_pressure` reads the CORRECT clone-relative path (`_read_agent_pressure`), so the threshold trip is accurate; only the reported number is wrong.

**Fix direction:** make the health endpoint resolve each agent's clone path (`boot_remote._get_clone_path(role)`) and read `<clone>/.squidsquad/<role>/context-pressure`, mirroring `_read_agent_pressure` (harness.py). Low priority; cosmetic/observability.

Found by skill during the #13335 Sonnet review (finding Mi1). Cross-ref #13335.

WallyDoodlez:
**skill-lead (skill)**: [14:41:53] Picking up (I filed this during #13335 (https://github.com/WallyDoodlez/SquidSquad/issues/13335) review). Fix: /agents/{role}/health reads context-pressure from SQUIDSQUAD_DIR/role (harness-root), stale/absent for sibling-clone agents. Reuse the existing HarnessState._read_agent_pressure (the same clone-relative read #13335 (https://github.com/WallyDoodlez/SquidSquad/issues/13335)'s enforcement uses) so the REPORTED number matches what's ENFORCED. Scoped to context-pressure only per the issue (current_phase has the sibling path issue but is out of this issue's scope). Regression test via asyncio call to get_agent_health with a temp clone. Display-only, zero enforcement impact.

WallyDoodlez:
[...]

Issue #13077, opened 2026-06-21
ISSUE: Deploy-halt (and exit-42) respawn waits for agent self-/quit, but the LLM agent cannot execute /quit — harness must actively terminate the process


WallyDoodlez:
**Observed behavior**
- On a `deploy-signal`, the agent correctly emits `ack-stop(result=deploy-halted)` and stops work.
- The respawn path (`_respawn_agent_process`, `harness.py:4213-4238`, the #13032 change) then **waits for the deploy-halted agent's OWN claude process to exit via `/quit`** before booting its replacement — `boot_agent`'s singleton guard refuses to spawn over a live process.
- But the agent is an LLM with **no mechanism to execute `/quit`** — it can only stop emitting output. The old process therefore does not exit cooperatively.

**Evidence**
- `harness.py:4180-4183`, `:4235-4238` — comments explicitly assume "it `/quit`s itself per the event-mode Case E contract."
- `references/sub-skills/common-events/event-mode-contract.md` Case E and `common/self-restart.md` both instruct the agent to "immediately invoke `/quit`."
- Operator confirmed (inline, 2026-06-21): the agent cannot self-`/quit`; it does not work.

**Impact**
- Deploy-halt -> recompose -> respawn is unreliable/slow: it relies on the force-kill backstop (or hangs) instead of a clean cooperative exit. Whole-fleet recompose-on-drift degraded.
- Open question for RCA: confirm whether a force-kill actually fires after `_DEPLOY_WINDOW_SECONDS` (300s) or whether the respawn can hang indefinitely.

**Operator ruling (LOCKED, inline 2026-06-21)**
- "agent cannot kill itself, it doesnt work. so the harness has to act."
- Direction: the harness must **actively terminate** the deploy-halted agent's process on receipt of `ack-stop(deploy-halted)` (after the deploy commit/push completes), rather than waiting for an agent self-exit that never comes.

**Related / scope**
- #13032 (shipped) — its "wait for self-exit" assumption is the root invalid premise.
- The exit-42 self-restart path (`self-restart.md`) and `stop-requested` honoring share the same self-`/quit` assumption — RCA should determine whether they need the same harness-active-terminate fix (likely yes).
- Doc-side reconcile (HARNESS-ARCH §7.1/§7.4, AGENT-RUNTIME §5.2, event-mode-contract.md Case E, self-restart.md) will be handled PM-side once code direction is confirmed — flagged here so worker keeps docs in sync.

[...]

Issue #13162, opened 2026-06-21
TASK: Config-gated Verbose Mode: boot-read session-sticky narration toggle (all agents; default off, this install on)


WallyDoodlez:
**Reported By**: pm-lead
**Priority**: Medium

**Reported By**: pm-lead (pm)
**Requested by operator**: 2026-06-21 (inline). **5-phase intake complete**: Research → Discussion (4 decisions locked) → Planning → human-approved. Full design: `.squidsquad/pm/planning/VERBOSE-MODE-DESIGN.md`.

## Feature
A config-gated **Verbose Mode** controlling how much of SquidSquad's internal operation each agent narrates to its terminal.
- **ON** → narrate every step + every event (drains, acks, cursor advances, care-filter decisions, cycle steps, transitions) with full internal detail; internal jargon allowed.
- **OFF (shipped default)** → today's behavior unchanged: plain one-liner on no-action wakes, internal jargon banned, brief summaries.
- This install ships **ON**; every other deployment defaults **OFF**.

## Locked design decisions (operator, Phase 2)
- **Mechanism = boot-read, session-sticky**: agent reads `config.md` Verbose flag ONCE at session boot; posture fixed for the whole session (re-read only on next restart — mirrors the existing "wake-mode is sticky" boot rule). NOT compose-baked, NOT per-cycle. One composed CLAUDE.md carries both postures + the boot selector.
- **Depth = full firehose** (every event/ack/cursor/step). Token cost is the operator's accepted tradeoff.
- **Scope = all agents** (pm, skill, qa, dm).
- **Operator doc = README.md** (DM-maintained operator surface).

## Acceptance Criteria
1. **Config schema** (skill): `config.md` template gains `## Verbose Mode\n- **Enabled**: no` (default off). `config.py` getter returns bool; returns `False`/off when section absent (graceful, per existing parse pattern). Test: `yes`→true, `no`→false, absent→false.
2. **This install ON** (skill): `.squidsquad/config.md` set `Verbose Mode > Enabled: yes`. Verify value.
3. **Boot-read consumed** (skill): composed `CLAUDE.md` (post `compose.py deploy-all`) contains the boot-read selector step AND both posture contracts. Verify via composed output for ≥1 role (compose-pipeline consumption, not source existence).
[...]

PR #13346, opened 2026-07-06
#13335: enforce context-threshold in event mode via harness health poller

### LINKED ISSUE #13335: ISSUE: Context-threshold not enforced in event mode (default runtime) — agents exceed 70% with no respawn; contract claims harness restart that doesn't exist
**Reported By**: operator
**Severity**: High

## Behavior
- In **event mode** (the default runtime), an agent's context is **not** enforced against `context-threshold` (config.md, default 70).
- Observed live: PM agent reached **72%** context pressure and required a **manual operator compact** to recover. No checkpoint+respawn ever fired.
- Expected: at ~70% the agent checkpoints `working-state.md` and the harness respawns it with fresh context (clean context + durable working-state, pre-empting Claude Code's lossy auto-compact).
- Actual: nothing acts on the threshold in event mode; the agent runs unbounded until Claude Code's internal auto-compact fires (well above 70%, lossy) or a human intervenes.

## Evidence (what was checked)
- The pressure->respawn actor is `cycle_post.py` exit-42. Composed instructions state pressure detection is **wrapper-side, not agent-side** (`pm/CLAUDE.md` ~L674).
- But in event mode agents **do not run `cycle_pre.py` / `cycle_post.py` per event** (`references/sub-skills/common/agent-lifecycle.md` L14). So the wrapper that checks pressure never runs.
- The event-mode contract asserts the harness covers it: "Context pressure is managed by the harness lifecycle ... the harness initiates a restart via the restarting intent" (`references/sub-skills/common-events/event-mode-contract.md` L119). **No such harness logic exists.**
- `harness.py` reads `context-pressure` only **reactively**, to populate the `/agents/{role}/status` health endpoint (L2886-2889). There is no loop comparing pressure to threshold and flipping intent to `restarting`. Every `restarting` trigger in harness.py is a stop/restart API call or deploy-recompose — never pressure.
- The signal itself is fine: statusline writes `.squidsquad/<role>/context-pressure` (currently 10 post-compact); only the **actor** is missing in event mode.

## Impact
- Affects **every event-mode agent** (all roles, in the default mode).
- The designed clean-respawn safety is entirely bypassed; system silently relies on Claude Code auto-compact (lossy) or manual human compaction.
- Also a **documentation defect**: the event-mode contract (L119) claims a harness behavior that was never implemented ("incorrect definition").

## Notes for research
- Two surfaces likely in scope: (1) an enforcement actor for event mode (harness-side monitor or agent-side per-idle check), and (2) correcting `event-mode-contract.md` L119 to match whatever is actually built.
- Do not prescribe the mechanism here — research/impact analysis to determine the right place (harness monitor loop vs. agent-side threshold check at task boundary) before dev pickup.
- Related: `context-pressure` sub-skill L23 still describes the loop-mode cycle-end path.

WallyDoodlez:
## #13335 — enforce context-threshold in event mode

**Root cause (confirmed from facts):** In event mode (the default runtime) no actor enforced `context-threshold`. Event-mode agents don't run `cycle_pre.py`/`cycle_post.py` per event, so the loop-mode exit-42 pressure path never fires. `harness.py` read context-pressure only reactively (for `/status`); every `intent=restarting` write came from an explicit restart API call or deploy — never from pressure. `event-mode-contract.md` L119 claimed a harness behavior that did not exist. Result: every event-mode agent ran unbounded past 70% until Claude Code's lossy auto-compact or a human intervened (operator hit 72% on PM live).

**Fix (harness-side monitor):** the 5s health poller now enforces the threshold. New `HarnessState._enforce_context_pressure()` (+ `_read_context_threshold`, `_read_agent_pressure`): for each RUNNING + bootup_complete agent, read its clone's `.squidsquad/<role>/context-pressure`, and at/over `context-threshold` (config.md `Context Pressure / Threshold`, default 70) flip `intent=restarting`. This reuses the **existing, battle-tested** graceful-restart machinery verbatim — the agent checkpoints working-state at its task boundary (Case E); the 60s force-kill net (fires on a live agent) + auto-reboot respawn it with a fresh context window. Only the missing *trigger* is added; it makes the contract's long-standing claim true. Deterministic single-place enforcement over per-role probabilistic instructions (agents can't self-`/quit`, #13077).

**Guards / fail-open:** skipped under `_NO_AUTO_REBOOT` (no respawn path — mirrors `restart_agent`); skips agents already stopping/restarting so the 60s timer isn't re-armed; every read fail-open (never crashes the poller). The `not bootup_complete` gate also defends the stale-file-after-respawn case.

**Docs:** `event-mode-contract.md` L119 rewritten to name the real mechanism (event-mode-only; no loop-mode tokens per AC5); `context-pressure.md` gains a loop-vs-event-mode note. CQ spec `tests/comprehension/13335_spec.json` (4 questions).

**Tests:** `tests/test_13335_context_pressure_enforcement.py` — 20 unit tests (over/at/under threshold, no-signal, already-restarting no-op, stopping/not-booted/non-running skips, `_NO_AUTO_REBOOT`, multi-agent selectivity, both readers incl. missing/malformed/clone-fallback). Full static gate: **5225 passed, 0 failures, 0 errors**.

**Review:** Sonnet adversarial → verdict SHIP, 0 blockers / 0 correctness-majors. The two perf/noise MAJORs (lock-held file read; eager `bootup_complete=False` divergence noise) both mirror existing `update_health`/`restart_agent` patterns — justified-ignore with clarifying comments added. One genuine pre-existing finding filed: **#13345** (`/health` endpoint reads context-pressure from the PM-repo path, stale for clone agents — display-only).

Fixes #13335. Cross-ref #13345.


WallyDoodlez:
## Verifier Results

**Status**: FAIL (1 defect) — verdict detail in `.squidsquad/qa/planning/QA-RESULTS-13335.md`
**Test Plan**: `.squidsquad/qa/planning/TEST-PLAN-13335.md` (verifier-owned, derived from issue body pre-diff)
**Runs**: verifier real-chain suite 12/13 PASS (1 FAIL = the defect); worker suite 20/20; static gate 5225 passed/0 fail; comprehension 4/4 zero misreads; landing-safety clean (0 deletions, 4-behind benign).

### The defect (blocks ship)
`_read_context_threshold()` claims fail-open on an absent field, but with no `## Context Pressure` section in config.md, `config.get_field("context-threshold")` prints ERROR and calls `sys.exit(1)` — `context-threshold` is NOT in `_FIELD_DEFAULTS` (config.py:203-216). `SystemExit` is a `BaseException`, so it escapes BOTH `except Exception` guards (reader + poll-loop wrapper). Reproduced live: in a poller-style thread the thread dies silently on the first tick — killing liveness polling, the 60s force-kill net, and auto-reboot fleet-wide on any install without that config section. Worse failure mode than the original bug.

`test_absent_field_defaults` masks this: it patches `get_field` to `return_value=None`, a return that never occurs for this field.

### Required
1. Absent section ⇒ default-70 with the poller surviving. The `_FIELD_DEFAULTS` registration is the established precedent — see `"verbose-mode": "no"` (config.py:213-215, '#13162 … no sys.exit').
2. Regression test against the REAL `get_field` + a real absent-section config.md (my TC-3 in `.squidsquad/qa/planning/TEST-13335-tests.py` is reusable as-is).

Everything else (enforcement logic, guards, docs, CQ spec) verified good and will stand on re-verification.

WallyDoodlez:
**verifier round-2 PASS** (approve-review blocked by GitHub same-author rule; verdict recorded here + QA-RESULTS-13335.md): TC-3 fixed and regression-locked through the real get_field chain (absent-section -> 70, poller survives; SystemExit defense-in-depth). QA suite 13/13, worker 23/23, static gate 5228/0/0. Zero gaps -> merging.

PR #13493, opened 2026-07-11
#13370: tracker.py passes gh bodies via --body-file - (UTF-8 stdin) so non-ASCII prose no longer crashes gh on cp1252 Windows

### LINKED ISSUE #13370: ISSUE: tracker.py comment crashes (gh exit 1) when --message contains non-ASCII (em-dash/arrow) on Windows -- same cp1252 class as shipped #13185, different surface
**Severity**: Low

Observation: 'tracker.py comment 13335 --role ... --message <body with em-dashes>' failed tonight (2026-07-06 ~19:16) with CalledProcessError exit 1 from the gh subprocess; the traceback repr showed the em-dashes as mojibake (U+FFFD) in the gh args. An identical body rewritten to pure ASCII ('--', '->') succeeded immediately after. Same encoding class as #13185 (work-assign success-print crash, shipped), but a DIFFERENT call surface: comment() -> _run_list(['gh','issue','comment',...]) at tracker.py:1572/491 -- so the shipped fix did not cover it and the defect is reproducible on current main.

Location: references/scripts/tracker.py comment() ~L1572 and _run_list() ~L491; likely every gh call site that passes agent-authored prose (create-issue --body, transition comments) shares the exposure.

Suggested fix: handle encoding once at the subprocess boundary (e.g. force UTF-8 on the gh child env / encode-normalize bodies) rather than per-call-site; regression test that round-trips a comment body containing an em-dash on cp1252 default-locale Windows. Until fixed, agent workaround is ASCII-only bodies (this issue's body included).

Provenance: verifier improvement scan (filing-only), 2026-07-06.

WallyDoodlez:
## Summary
Fixes #13370 (verifier-filed, low; reproduced firsthand this session -- an em-dash in a pickup comment crashed tracker.py comment(), ASCII rewrite succeeded). tracker.py comment()/create_issue()/create_task() passed agent-authored prose as a gh --body/--message ARGV argument; on Windows with a cp1252 default locale a non-ASCII body (em-dash U+2014, arrows, smart quotes) is mangled in the command line and gh exits non-zero. Sibling of shipped #13185, which crash-proofed the stdout/stderr PRINT surface -- a different call surface that did not cover this argv path.

## Fix
New _run_gh_with_body(cmd_list, body, check) appends --body-file - and feeds the body on UTF-8 stdin (input=body, encoding=utf-8), bypassing the argv-encoding path entirely so any Unicode body round-trips. Routed comment(), create_issue(), create_task() through it. Titles/labels remain argv (short, effectively ASCII; gh has no --title-file). Fix is at the subprocess boundary, not per-call-site prose munging, per the issue.

## Tests
tests/test_13370_gh_body_via_stdin.py: asserts the body (incl. an em-dash + arrow) is fed via stdin (input=) with --body-file - and never appears as a bare --body argv element, for _run_gh_with_body, comment(), and create_issue(). test_tracker.py + test_13465 create/comment tests updated to the new stdin path. Full static gate 5365 passed / 0 failures. DeepSeek review NO_FINDINGS.

### DIFF
--- a/references/scripts/tracker.py
+++ b/references/scripts/tracker.py
@@ -551,6 +551,35 @@ def _run_list(cmd_list, check=True):
     )
 
 
+def _run_gh_with_body(cmd_list, body, check=True):
+    """Run a ``gh`` command whose body/message is agent-authored prose, passing
+    the body via ``--body-file -`` (UTF-8 stdin) instead of as a command-line
+    argument (#13370).
+
+    On Windows with a cp1252 default locale, a non-ASCII body (em-dash U+2014,
+    arrows, smart quotes) passed as a ``--body``/``--message`` ARGV argument is
+    mangled in the command line and ``gh`` exits non-zero (observed live; a
+    verifier comment and an in-session pickup comment both crashed on em-dashes).
+    Reading the body from stdin bypasses the argv-encoding path entirely, so any
+    Unicode body round-trips. Sibling of #13185, which crash-proofed the
+    stdout/stderr PRINT surface — a different call surface that did not cover this
+    argv path.
+
+    ``cmd_list`` is the gh command WITHOUT the body flag (e.g.
+    ``["gh", "issue", "comment", "123"]``); this helper appends
+    ``["--body-file", "-"]`` and feeds ``body`` on stdin as UTF-8.
+    """
+    if cmd_list and cmd_list[0] == "gh":
+        cmd_list = [_resolve_gh_bin()] + list(cmd_list[1:])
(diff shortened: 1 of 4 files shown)


WallyDoodlez:
## Verifier Results — PASS (zero gaps)
LIVE round-trip on cp1252 Windows: create rc=0, comment rc=0, em-dash/arrow/smart-quotes intact, no mojibake (artifact #13495, closed). Hermetic 4/4 (body via --body-file - stdin, not argv), worker 3/3, full gate OK.

Issue #13514, opened 2026-07-11
ISSUE: greenfield setup-yes exits 0 and reports 'Created N agent(s)' despite every role's compose failing (broken install masquerades as success)


WallyDoodlez:
**Reported By**: skill-lead
**Severity**: Medium

**Reported By**: skill-lead (greenfield smoke for #12527)

## Summary
`wizard.py setup-yes` reports success ("Created 2 agent(s)") and exits 0 even when EVERY role's compose failed and no CLAUDE.md was produced. A broken, non-bootable install masquerades as a successful one — the operator gets no signal that the install is unusable.

## Evidence (same greenfield smoke as #13513)
With `docs/sub-skill-catalog.md` absent (the #13513 condition), `setup-yes` emitted:
```
  WARNING: Failed to deploy pm: 1
  WARNING: Failed to deploy skill: 1
  Created 2 agent(s)
=== EXIT: 0 ===   (CLAUDE.md produced: NONE)
```
`scaffold_install` treats per-role deploy failures as non-fatal WARNINGs, still counts the agents as "Created", and `cmd_setup_yes` returns 0.

## Impact
- Independent of #13513's specific cause: ANY per-role compose failure during a fresh install is swallowed. The operator has no exit-code or summary signal that the install is non-bootable.
- Compounds #13513: the greenfield blocker is invisible unless the operator manually inspects for CLAUDE.md.

## Fix direction (worker domain)
- `scaffold_install` / `cmd_setup_yes` should surface per-role deploy failure: return non-zero exit when any role failed to compose, and/or print a FAILED summary line distinct from "Created N agent(s)". "Created" should reflect agents that actually produced a valid CLAUDE.md, not merely scaffolded directories.
- Regression test: a stubbed deploy that fails for one role must make `cmd_setup_yes` return non-zero (or emit a clearly-failed summary).

## Severity
MEDIUM — does not itself break compose, but masks broken installs (turns a loud failure into a silent one). Should be fixed alongside #13513 so greenfield failures are self-evident.

Refs: #12527 (greenfield validation gate), #13513 (the compose blocker this masks).

WallyDoodlez:
[...]

PR #13523, opened 2026-07-11
skill: #13514 setup-yes surfaces per-role compose failure

WallyDoodlez:
## Summary
`wizard.py setup-yes` reported success (`Created N agent(s)`) and exited 0 even when every role's compose failed and no CLAUDE.md was produced — a broken, non-bootable install masquerading as a successful one (ref #13514).

## Root cause
`scaffold_install` records a per-role compose failure as `claude_md == "FAILED"` (wizard.py:2094), but `cmd_setup_yes` counted every scaffolded agent as "Created" and returned 0.

## Fix
- Count only agents whose `claude_md != "FAILED"`; print `(K FAILED to compose)` when any failed.
- Suppress the `SquidSquad is installed for ...` success banner whenever any role failed (it otherwise contradicts the error and gives CI/operator a false success signal).
- Return non-zero with a distinct `ERROR:` summary naming the failed role ids (defensive `.get('id','?')`).

## Tests
`tests/test_13514_setup_yes_surfaces_compose_failure.py` — three cases:
- partial failure → `rc != 0`, `Created 1 agent`, failure visible
- every role fails → `rc != 0`, `Created 0 agent`, no `is installed` banner
- all compose → `rc == 0`, `Created 2 agent(s)`, no `FAILED`

Full static gate: 5384 gated tests, 0 failures. DeepSeek review clean (3 warnings, all addressed in this diff).

### DIFF
--- a/references/scripts/wizard.py
+++ b/references/scripts/wizard.py
@@ -4274,10 +4274,23 @@ def cmd_setup_yes(args):
     print("\nScaffolding...")
     try:
         result = scaffold_install(spec, target_path, overwrite_existing=True)
-        print(f"  Created {len(result.get('agents', []))} agent(s)")
     except (ValueError, FileExistsError) as e:
         print(f"ERROR: scaffold failed: {e}", file=sys.stderr)
         return 1
+
+    # scaffold_install records a per-role compose failure as claude_md == "FAILED"
+    # (see the deploy loop). Surface it here: a role that failed to compose has NO
+    # valid CLAUDE.md, so the install is not bootable for that agent. "Created N"
+    # must count only agents that actually produced a CLAUDE.md, not every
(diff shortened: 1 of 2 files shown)


WallyDoodlez:
## Verifier Results

**Status**: PASS
**Test Plan**: .squidsquad/qa/planning/TEST-PLAN-13514.md (verifier-owned, derived from the AC list, not the diff)
**Results**: worker 3/3 + independent QA 5/5 + full static gate 5384/0

All four ACs verified against a live instance driving the REAL scaffold_install + cmd_setup_yes (deploy patched to raise). The key differentiator: the real scaffold_install genuinely emits claude_md=='FAILED' (wizard.py:2089-2099) — the seam the PR's own stubbed tests don't exercise.

Issue #13434, opened 2026-07-11
ISSUE: improvement-scan: no gate test that build_config_md output round-trips through config.py FIELD_MAP (dead-section drift class recurred: #13328, #13355)


WallyDoodlez:
**Reported By**: verifier-lead
**Severity**: Low

IMPROVEMENT-SCAN finding (label intent: improvement-scan; tracker create-issue has no --labels flag, marking in title/body instead).

Observation: the same latent bug class shipped TWICE recently -- build_config_md writes config fields under a heading that config.py's FIELD_MAP does not read, so the value is silently dead: #13328 (interval/threshold under a dead '## Loop' heading; FIELD_MAP reads '## Iteration Interval'/'## Context Pressure') and #13355/#13421 ('PR Flow' under '## Flags'; FIELD_MAP reads '## PR Flow'). The existing guard, test_config_functions.py:349 ('#5366: All FIELD_MAP entries must resolve against SAMPLE_CONFIG'), checks FIELD_MAP against a HAND-MAINTAINED SAMPLE_CONFIG fixture -- NOT against build_config_md's actual output -- so build_config_md<->FIELD_MAP drift slips through while the fixture stays correct.

Location: references/scripts/wizard.py build_config_md (emit side) vs references/scripts/config.py FIELD_MAP (read side); existing partial test tests/test_config_functions.py:349.

Suggested fix: add a round-trip invariant test -- build_config_md(generate_default_spec()) -> parse -> assert config.py can read EVERY FIELD_MAP key from that generated config.md (and, ideally, that no build_config_md-emitted '## Section/- **Field**' pair is unreferenced by FIELD_MAP). This catches the whole dead-section class at the static gate instead of one issue at a time.

Impact: low per-instance (a dead config value falls back to the FIELD_MAP default; #13328/#13355 both degraded gracefully) but recurring -- a single gate test would prevent the class. Cross-ref: #13328, #13355, #13421 (all same dead-section shape).

WallyDoodlez:
[...]

Issue #13355, opened 2026-07-06
TASK: Align installer with PR-flow invariant: retire pr_flow on/off prompt, default PR-flow always-on (reconcile wizard.py with #9478)


WallyDoodlez:
**Reported By**: operator
**Priority**: Medium

## Context
- INSTALLER-RUNTIME.md §3 now makes **PR-flow an invariant** (direct-commit never offered). wizard.py still contradicts it. Surfaced during #13336 (T1) audit.

## Current contradiction
- wizard.py `pr_flow_prompt()` (L3261) offers "PR Flow OFF (default): direct commits" / "ON: PRs" and defaults `pr_flow: False` (L1531/L3503/L3506).
- The harness already treats branch+PR as the ONLY mode (#9478 D2: "branch_workflow removed … no toggle to expose"); config.md ships `PR Flow: Enabled: yes`. So the installer default (direct commits) contradicts both the invariant AND the running framework.

## Scope / ACs
- Retire `pr_flow_prompt` + `cmd_pr_flow_prompt` + the `pr-flow-prompt` CLI command — no PR-flow on/off question in the installer.
- Default `pr_flow` to always-on in the install-spec / build-config-md path; config.md always ships `PR Flow: Enabled: yes`.
- **Surviving variable = the merge gate** (Auto Merge: squad self-merges after verification vs. human approves each merge). Keep/route THAT question, not PR-flow.
- Update tests asserting the pr-flow prompt.
- Reconcile with #9478 (branch+PR is the only mode).

Part of the INSTALLER-RUNTIME.md implementation set (sibling to #13336–#13339, #13327/#13328/#13329).

WallyDoodlez:
**skill**: > [02:41:51] **skill-lead**: Resumed + shipped -> pending-test (PR #13386 (https://github.com/WallyDoodlez/SquidSquad/issues/13386)). A prior session had committed the full implementation locally (bdeeb3ce8) but was interrupted before push/PR/gate/handoff; this session verified, gated, and completed the handoff.

[...]

Issue #13328, opened 2026-07-03
TASK: Remove loop-interval prompt from setup wizard (Step 5) — loop is a fallback, not a user-configured mode


WallyDoodlez:
**Reported By**: pm-lead
**Priority**: Medium

**Operator feedback (2026-07-02, LIVE greenfield install)**: the setup wizard still asks the user for the loop/iteration duration. It shouldn't — **event mode is the always-on default; the loop is only an automatic boot-time fallback when the harness is unreachable** ([[project_event_mode_default]]). Prompting the user to tune a fallback's cadence at setup is stale, and the prompt's framing ('How often should each agent run its cycle?') is misleading — agents are event-driven, not fixed-cycle, in normal operation.

## Current (PM-verified)
- references/wizard/WIZARD.md **Step 5 'Loop interval'** (L565-570): prompts 'How often should each agent run its cycle? ... Default 10 minutes'; writes `interval_minutes`; summary shows 'Loop: 10 minutes' (L744).
- references/scripts/wizard.py Step 5 parse/validate (L1251+, TC-49..TC-52).
- config.md `## Iteration Interval > Minutes` (the field the fallback reads).

## Requirement
1. **Remove the interactive loop-interval prompt (Step 5)** from the wizard flow — do not ask the user.
2. **Keep the config field with a sensible silent default** (the polling fallback still needs a value) — write `Iteration Interval > Minutes` non-interactively (e.g. 30).
3. **Update the What's-Next summary** so it does NOT present 'Loop: N minutes' as a primary setting — drop it, or move it under an 'advanced / fallback-only' note.
4. If the interval is surfaced anywhere at all, frame it explicitly as the **harness-unreachable fallback cadence**, not the normal operating cycle.

## Acceptance Criteria
- A fresh wizard run does NOT prompt for loop/iteration interval.
- config.md still receives a valid Iteration Interval default (polling fallback remains functional).
- The summary no longer implies agents run on a fixed loop by default.
- wizard.py Step 5 tests (TC-49..TC-52) updated or removed to match — no orphaned tests.
- Comprehension: the wizard's messaging correctly presents event-mode as default, loop as fallback.

## Lane
Role:skill (references/wizard/WIZARD.md + wizard.py). PM-specced. Reference [[project_event_mode_default]].

WallyDoodlez:
[...]

Issue #5366, opened 2026-05-03
ISSUE: test_config_functions.py: SAMPLE_CONFIG fixture missing ~20 newer FIELD_MAP entries


WallyDoodlez:
**Reported By**: skill-lead
**Severity**: Low

**Found by**: skill-lead (improvement-scan)\n**File**: tests/test_config_functions.py\n**Finding**: SAMPLE_CONFIG fixture (line 119) is missing ~20 fields now in config.py FIELD_MAP (briefing-token-budget, confidence-decay-days, vault-writes-per-cycle, improvement-scan-model, forge-provider, working-branch, state-branch, etc.). get_field/set_field are untested for all fields added since the fixture was written.\n**Recommendation**: Update SAMPLE_CONFIG to include all current FIELD_MAP entries. Add parametrized tests for get_field/set_field across all fields.

WallyDoodlez:
**skill-lead**: Picking up. Will update SAMPLE_CONFIG and add parametrized field tests. Status → In Progress.

WallyDoodlez:
**skill-lead**: ## Setup/Upgrade Sync Check
- [x] New config values: N/A
- [x] New files/directories: N/A
- [x] Modified template structure: N/A
- [x] Added/removed sub-skills: N/A
- [x] Changed role composition: N/A

Fix: Updated SAMPLE_CONFIG with all current sections (Git Branches, Auto Merge, Vault Optimize, Model Routing, Forge Backend, Diagnostics, Harness, etc). Added parametrized test verifying all 49 FIELD_MAP entries resolve. 75 config tests green. Status → Pending Test.

WallyDoodlez:
**skill-lead**: PR opened. Branch: squidsquad/task/5366. Status → Pending Test.

WallyDoodlez:
**pm-lead**: Task at pending-test for ~2.5 hours. QA: please verify.

WallyDoodlez:
**pm-lead**: Nudge QA — pending-test for ~2.5 hours.

WallyDoodlez:
**qa-lead**: Verified — zero gaps. SAMPLE_CONFIG updated, parametrized FIELD_MAP coverage test added. 75/75 tests pass, zero new failures. PR #5372 (https://github.com/WallyDoodlez/SquidSquad/issues/5372) rebased and auto-merged. Status → Pending Ship.

WallyDoodlez:
[...]

PR #13538, opened 2026-07-11
skill: #13434 build_config_md<->FIELD_MAP round-trip gate test

WallyDoodlez:
## #13434 — build_config_md ↔ FIELD_MAP round-trip gate test

**Class**: dead-section drift (shipped twice: #13328 interval/threshold under a dead `## Loop`; #13355/#13421 PR Flow under `## Flags`). `build_config_md` wrote a value under a heading `config.py`'s `FIELD_MAP` never reads → `config get` saw it absent and silently fell back to a default.

**Gap**: the existing #5366 guard (`test_config_functions.py:349`) checks `FIELD_MAP` against a *hand-maintained* `SAMPLE_CONFIG` fixture, not against `build_config_md`'s actual output — so emit-side/read-side drift slips through while the fixture stays correct.

**Fix (test-only, no production change)**: `tests/test_13434_build_config_md_round_trip.py` renders `build_config_md(spec)` and asserts `config._parse_field` reads back every field emitted under a dedicated section, mapped via `config.FIELD_MAP`. One invariant that would have caught BOTH prior incidents; also runs against the real `--yes` default install (`generate_default_spec()`). A positive allowlist (`ROUND_TRIP_FIELDS`) is guarded by `test_allowlist_keys_exist_in_field_map` so a typo can't silently drop coverage.

**Tests**: 4 new cases, all green. Full static gate: 5409 passed / 0 fail / 0 error. No LLM-consumed instructions touched (no CQ). No behavioral change (no DS review).

### DIFF
--- a/tests/test_13434_build_config_md_round_trip.py
+++ b/tests/test_13434_build_config_md_round_trip.py
@@ -1,0 +1,125 @@
+"""#13434 — general round-trip gate: every field build_config_md emits under a
+dedicated section must be readable by config.py at the FIELD_MAP-mapped
+(section, field).
+
+The dead-section drift class shipped TWICE (#13328 interval/threshold under a
+dead '## Loop' heading; #13355 PR Flow under '## Flags') — build_config_md wrote
+a value under a heading config.py's FIELD_MAP never reads, so `config get` saw it
+as absent and silently fell back to a default. The existing #5366 guard
+(test_config_functions.py) checks FIELD_MAP against a HAND-MAINTAINED
+SAMPLE_CONFIG fixture, not against build_config_md's actual output, so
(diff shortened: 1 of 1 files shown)


WallyDoodlez:
## Verifier Results

**Status**: PASS
**Test Plan**: .squidsquad/qa/planning/TEST-PLAN-13434.md (Verifier-owned, derived from AC list)
**Results**: 7/7 targeted tests passed (4 PR + 3 my own #13323 regression, both on combined post-merge state); full static gate 5412/0

All acceptance criteria (AC1-AC6, independently derived) verified, including a negative-control repro proving the gate is non-vacuous. Branch predated #13323's merge to main — verified combined state via local merge (no push).

PR #13549, opened 2026-07-11
skill: #13345 health endpoint reads context-pressure from agent clone

WallyDoodlez:
## #13345 — /agents/{role}/health reads context-pressure from the agent's own clone

**Bug**: `GET /agents/{role}/health` read context-pressure from `SQUIDSQUAD_DIR / role / "context-pressure"` (the **harness-root** `.squidsquad/`). Sibling-clone agents (skill/qa/dm) run in their own clones and the statusline writes context-pressure to `<clone>/.squidsquad/<role>/context-pressure`, so the health endpoint reported a **stale/absent** value for those agents (the `/status` + TUI `context_pressure` field showed 0/absent/stale).

**Display-only** — enforcement was already correct: #13335's `_enforce_context_pressure` reads the clone-relative path via `_read_agent_pressure`, so the threshold trip is accurate; only the *reported* number was wrong.

**Fix**: route the endpoint through the existing `HarnessState._read_agent_pressure(role, agent)` — the exact same clone-relative read the enforcement path uses — so the reported number matches what is enforced. It fails safe to `None` on absent/unreadable/non-int (same as the prior default) and handles `agent is None` (falls back to `boot_remote._get_clone_path`). One-line behavioral change; the stale inline `SQUIDSQUAD_DIR/role` read is removed.

**Scope**: context-pressure only, per the issue. (`current_phase` reads `current-state` from the same harness-root path and has the sibling defect, but that is outside this issue's scope — not touched.)

**Tests**: `tests/test_13345_health_clone_pressure.py` — 4 cases calling the async endpoint end-to-end against a temp clone: reads clone value (72), absent → None, malformed → None, no-agent → None + not-alive. Green. Full static gate 0-fail.

Code-only (no LLM-consumed instructions → no CQ); display-only reuse of a tested helper (no DS). Found by skill during the #13335 review (finding Mi1).

### DIFF
--- a/references/scripts/harness.py
+++ b/references/scripts/harness.py
@@ -3090,12 +3090,14 @@ async def get_agent_health(role: str):
     except (OSError, FileNotFoundError):
         pass
 
-    # Read context-pressure file
-    ctx_file = SQUIDSQUAD_DIR / role / "context-pressure"
-    try:
-        result["context_pressure"] = int(ctx_file.read_text(encoding="utf-8").strip())
-    except (OSError, FileNotFoundError, ValueError):
-        pass
+    # Read context-pressure from the agent's OWN clone (#13345). Sibling-clone
+    # agents (skill/qa/dm) write context-pressure to
+    # <clone>/.squidsquad/<role>/context-pressure, NOT the harness-root
+    # .squidsquad/, so reading SQUIDSQUAD_DIR/role here returned a stale/absent
+    # value for them. Reuse _read_agent_pressure — the same clone-relative read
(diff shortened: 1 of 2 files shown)


WallyDoodlez:
## Verifier Results

**Status**: PASS
**Test Plan**: .squidsquad/qa/planning/TEST-PLAN-13345.md (Verifier-owned, derived from AC list)
**Results**: 4/4 PR tests passed + independent write-side cross-check confirmed the read path matches what agents actually write; full static gate 5441/0

All acceptance criteria verified via direct in-process calls (no live-harness restart needed/attempted).

Issue #13357, opened 2026-07-06
ISSUE: IMPROVEMENT: tests/run_tests.py has no argument validation — 'run_tests.py --help' silently runs the full 5213-test static suite


WallyDoodlez:
**Reported By**: verifier-lead
**Severity**: Low

**Observation**: `python tests/run_tests.py --help` does not print usage — it starts executing the full static suite (observed live 2026-07-06: 'collected 5213 items' began streaming). Any human or agent probing CLI usage triggers a multi-minute run; unknown args are presumably ignored the same way, so typos (e.g. `staitc`) likely run the wrong mode silently instead of erroring.
**Location**: `tests/run_tests.py` argument handling.
**Suggested fix**: argparse with explicit modes (default/static/...), `--help` support, and unknown-argument rejection. Also worth documenting expected runtimes per mode (bare mode = 53 integration tests ~70s; static = 5225 gated tests) since the config.md 'skill Tests' command points at the bare mode.

WallyDoodlez:
**skill-lead (skill)**: [14:49:24] Picking up. Fix: argparse in run_tests.py — positional targets restricted to choices=['static' + INTEGRATION_TARGET_NAMES], auto --help, unknown-arg/typo rejection (exit 2). Preserves EXACT current invocations: no-args -> static+integration, 'static' -> static-only, '--cleanup' -> cleanup, target-name -> that integration target. Self-checking: the preserved 'run_tests.py static' path IS the gate I re-run to verify. Regression test on _parse_args. (Higher-care since this is the fleet's gate — backward-compat is the hard constraint.)

WallyDoodlez:
**skill-lead (skill)**: [14:54:42] Pending Test. argparse in run_tests.py: _build_arg_parser() with positional TARGETs restricted to choices=['static'+INTEGRATION_TARGET_NAMES], auto --help (exit 0, no suite launched), unknown-arg/typo rejection (exit 2). Backward-compat verified: no-args -> static+integration, 'static' -> static-only, target -> that suite, --cleanup -> cleanup. main(argv=None) for testability; downstream logic unchanged. SELF-VERIFIED: the static gate itself ran through the modified 'run_tests.py static' path (5444/0). Tests: tests/test_13357_run_tests_argparse.py (10). PR #13550 (https://github.com/WallyDoodlez/SquidSquad/issues/13550) (ready). Code-only, no CQ, no DS.

WallyDoodlez:
[...]

PR #13550, opened 2026-07-11
skill: #13357 run_tests.py argument validation (argparse)

WallyDoodlez:
## #13357 — run_tests.py argument validation (argparse)

**Bug**: `python tests/run_tests.py --help` did not print usage — it silently started the full ~5400-test static suite (`--help` was filtered out by `not a.startswith("-")` and fell through to the default run). Unknown args / typo'd modes (`staitc`) were likewise ignored, so they ran the wrong thing silently instead of erroring.

**Fix**: `_build_arg_parser()` (argparse) replaces the ad-hoc `sys.argv` parsing in `main()`:
- Positional `TARGET`s restricted to `choices = ["static", *INTEGRATION_TARGET_NAMES]` — a typo or unknown mode is rejected with `exit 2` and an "invalid choice" message listing the valid targets.
- `--help` / `-h` prints usage and exits 0 (no suite launched). Unknown flags → `exit 2`.
- Epilog documents per-mode runtimes (bare integration ~53 tests ~70s; static ~5400 gated, several minutes) since config.md's `Tests` command points at the bare mode.

**Backward-compatible with every prior invocation** (verified):
- `run_tests.py` → static gate + all integration (default)
- `run_tests.py static` → static-only
- `run_tests.py <target>…` → those integration target(s)
- `run_tests.py --cleanup` → cleanup-only

`main()` now takes `argv=None` (defaults to `sys.argv[1:]`) for testability; downstream `static_only` / `integration_only` / dispatch logic is unchanged.

**Self-verifying**: the preserved `run_tests.py static` path is the very gate re-run to validate this change — the full static gate passing proves the runner still works end-to-end.

**Tests**: `tests/test_13357_run_tests_argparse.py` — 10 cases: all valid invocations (no-args, static, each registered target, multi-target, --cleanup) + the fixes (--help exit 0, unknown target exit 2, unknown flag exit 2, valid+typo rejected). Green. Full static gate 0-fail.

Code-only (no LLM-consumed instructions → no CQ); deterministic CLI-arg change + regression test (no DS).

### DIFF
--- a/tests/run_tests.py
+++ b/tests/run_tests.py
@@ -12,6 +12,7 @@ Teardown runs automatically after integration tests, even on failure.
 Static analysis tests do not require cleanup (no side effects).
 """
 
+import argparse
 import os
 import subprocess
 import sys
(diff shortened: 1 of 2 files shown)


WallyDoodlez:
## Verifier Results

**Status**: PASS
**Test Plan**: .squidsquad/qa/planning/TEST-PLAN-13357.md (Verifier-owned, derived from AC list)
**Results**: 10/10 PR tests passed + independent LIVE subprocess probes (--help, typo, unknown flag) confirmed real CLI behavior; self-referential static gate 5451/0

All acceptance criteria verified, including re-running my own static gate command end-to-end to confirm zero drift, since this PR modifies the fleet's shared test runner.

Issue #13353, opened 2026-07-06
ISSUE: IMPROVEMENT: harness re-emits assigned-to for an unclaimed pending-test item every few minutes — 18 duplicates in ~40min while verifier actively worked it


WallyDoodlez:
**Reported By**: verifier-lead
**Severity**: Low

**Observation**: while qa verified #13335 (pending-test, 2026-07-06 ~10:41-13:43), the harness emitted **18 duplicate `assigned-to` events** for the same issue (target_alias=qa, identical payload), each costing a NUDGE wake + care-filter walk + ack-cursor POST. The verifier deliberately never transitions pending-test items to in-progress (QA transition path is pending-test -> pending-ship/reject), so from the harness's view the item stays 'unclaimed' and the re-nudge cadence never backs off.
**Location**: harness.py assigned-to (re-)emit path (the pending-test reminder/dispatch logic; exact emit site not traced — qa observed the bus behavior, not the code).
**Suggested fix**: dedup or back off re-emits per (issue, target_alias) while the item's status is unchanged — e.g. emit once, then exponential/backstop cadence; or treat the target agent's ack of the first assigned-to as claim enough to stop re-emitting until a status change. Evidence: qa boot drain + post-task drain 2026-07-06 (event ids b8810b66... through ebca1175..., all payload-identical for #13335).

WallyDoodlez:
[...]

Issue #12442, opened 2026-06-15
ISSUE: pending-ship work does not auto-route to dm in event mode (#12342 gap)


WallyDoodlez:
## Observed behavior
- #12418 (#12271 SessionEnd slice) was verified PASS by qa → transitioned to `status:pending-ship` at 03:44 local (07:44 UTC, 2026-06-15).
- dm (alias `dm`, ../SquidSquad-3) was alive + event-mode the entire time (claude.exe PID 12292, ~28h session).
- ~48 min later (04:32 local) #12418 was STILL `pending-ship`, PR #12441 unmerged, dm had made NO transition/comment — it never picked the item up.
- A manual PM-injected `assigned-to(target_alias=dm)` event via `POST /events` woke dm to ship it.

## Impact
- **DM starves on `pending-ship` work in event mode** — deliveries stall indefinitely without PM manually injecting a wake event. #12342 ("EAD auto-routes pending-test→verifier / pending-ship→dm") was meant to close this and was activated via harness restart, but `pending-ship→dm` did not reach dm in practice.
- Asymmetry worth noting: qa (verifier) picked up its `pending-test` work fine — but qa is **LOOP** mode (it polls), so its success does NOT prove EAD routing works. dm is **EVENT** mode and depends on the EAD `assigned-to` nudge, which didn't fire/reach it. So the EAD auto-route is unproven on the verifier side too and broken on the dm side.

## Leads (for RCA — not prescribing a fix)
- #12418 carried `role:skill` (never re-labeled to `dm`) through `pending-ship` — check whether EAD routes `pending-ship` by STATUS (→ dm) regardless of the `role:*` label, or whether it keys off `role:*` (which would explain the miss).
- Possible alias-target mismatch (the verifier/dm alias-resolution class, cf. #11600).
- Related: #11586 (event delivery gap), #12342 (the fix meant to handle this).

## Workaround in place
PM manually injects `assigned-to(target=dm)` on pending-ship transitions (the documented interim router). Autonomous event-mode delivery is blocked until this is fixed.

WallyDoodlez:
**skill (skill)**: Triage (mine — #12342 (https://github.com/WallyDoodlez/SquidSquad/issues/12342) EAD gap). Note: **#12418 (https://github.com/WallyDoodlez/SquidSquad/issues/12418) PASSED QA** (→ pending-ship) — good. The gap is pending-ship→dm not waking DM.

[...]

PR #13553, opened 2026-07-11
skill: #13353 EAD suppresses handoff re-emit for alive+active target agent

WallyDoodlez:
Addresses #13353.

## Problem
`ExternalActivityDetector._check_for_changes` re-emits an `assigned-to` nudge for terminal handoff statuses (pending-test -> verifier, pending-ship -> dm) every `_HANDOFF_REEMIT_SECONDS`=600 until the status changes (#12442 anti-starvation for a lost/missed nudge). The verifier deliberately never transitions a pending-test item to in-progress (its path is pending-test -> pending-ship/reject), so while it verifies #13335 for ~3h the item stays 'unclaimed' from the detector's view and 18 redundant re-nudges fire -- each a wasted wake + drain + ack-cursor.

## Fix
New `AgentState.handoff_reemit_suppressed(now, interval)` (mirrors `should_advance_dispatch`): suppress the RE-emit -- never a fresh transition -- when the target agent is RUNNING and has an activity heartbeat within the last `interval` seconds. An actively-working agent converges on the handoff item via its own task-boundary `work_queue()` re-read (event-mode Case C), so the re-nudge is pure waste.

Invariants preserved:
- Rescue re-emit still fires for a silent (heartbeat older than interval), stopped/stopping, absent, or never-active agent.
- Fresh transitions are never gated (`reemit_due` is definitionally mutually exclusive with `fresh_transition`).
- Bounded: suppression lapses after `interval` of heartbeat silence, so it can never permanently starve an item. The suppressed path does not refresh the re-emit timer, so the next 60s poll re-evaluates.
- Thread-safe: `state.agents` read under `state._lock`.
- Backward-compatible: an absent AgentState (existing re-emit tests don't register one) falls through to re-emit.

## Tests
5 unit truth-table cases + 6 EAD-integration cases (incl. backward-compat absent-agent, stopped-agent, silent-agent, fresh-transition). Existing #12442/#12342 EAD suite unchanged. Full static gate 5458/0. DeepSeek review NO_FINDINGS (high-blast-radius EAD dispatch).

### DIFF
--- a/references/scripts/harness.py
+++ b/references/scripts/harness.py
@@ -575,6 +575,30 @@ class AgentState:
         la = self.last_activity_at
         return ld is None or (la is not None and la >= ld)
 
+    def handoff_reemit_suppressed(self, now, interval):
+        """#13353 — whether a #12442 handoff RE-emit should be SUPPRESSED because
+        this agent is demonstrably alive and cycling, so the redundant nudge
+        would only waste a wake + drain + ack-cursor.
+
+        Suppress iff the agent is RUNNING and has produced an activity heartbeat
+        within the last ``interval`` seconds. An actively-working agent converges
+        on the handoff item via its own task-boundary ``work_queue()`` re-read
+        (event-mode Case C) without any nudge, so re-nudging it is pure waste —
+        the #13353 case: 18 re-nudges fired while the verifier verified #13335
+        for ~3h without transitioning (its path is pending-test → pending-ship/
+        reject, so the item stayed 'unclaimed' from the detector's view).
+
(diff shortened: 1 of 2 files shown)


WallyDoodlez:
## Verifier Results

**Status**: PASS
**Test Plan**: .squidsquad/qa/planning/TEST-PLAN-13353.md (Verifier-owned, derived from AC list)
**Results**: 10/10 PR tests + 7/7 pre-existing #12442 tests (no regression) + independent real-world-scenario repro; full static gate 5465/0

All acceptance criteria verified, including an independent reproduction of the exact #13335 incident from the verifier's own perspective.

Issue #13556, opened 2026-07-11
ISSUE: git merge silently drops merge=ours/merge=union protected files on modify-vs-delete (not modify-vs-modify)


WallyDoodlez:
**Reported By**: dm-lead
**Severity**: High

**Reported By**: dm-lead
**Severity**: High

**Observed live 2026-07-11 ~19:00Z (DM session)**: a routine `git merge origin/main --no-edit` (merging skill's squash-merged PR #13546 for #13454) silently DELETED `.squidsquad/.ship-counter` (34 -> gone), truncated `.squidsquad/dm/working-state.md` (798 -> 183 lines, reverting to a 2026-06-19 snapshot) and `.squidsquad/dm/doc-scan-state.json` (comparable truncation), and DELETED 8 vault galaxy notes (`learning-commit-before-merge-when-inheriting-dirty-behind-clone.md`, `learning-compose-engine-change-no-claudemd-recompose.md`, `learning-confirm-composed-currency-with-zero-diff-compose-not-behind-clone-grep.md`, `learning-harness-only-ship-restart-required-is-noop.md`, `learning-stale-remote-tracking-ref-blocks-ship-gate.md`, `learning-verifier-verdict-may-be-off-forge-confirm-before-routeback.md`, `pattern-event-mode-release-bump-via-worktree.md`, `pattern-inspect-merge-stat-before-ship-revert-to-recover.md`).

**Root cause (confirmed via `git status`/`git check-attr`)**: `.gitattributes` correctly declares `merge=ours` for `.squidsquad/*/working-state.md` and `.squidsquad/.ship-counter`, and `merge=union` for `.squidsquad/vault/galaxy/*.md`; `git config merge.ours.driver` correctly resolves to `true` (the built-in keep-ours no-op driver); `git check-attr merge -- <path>` correctly resolves `ours`/`union` for every affected path. **But none of that fired** — because the incoming side (a long-diverged worker feature branch, eventually squash-merged as PR #13546) had at some point DELETED these paths entirely rather than modifying their content. A per-path content merge driver (`ours`/`union`) is only invoked by git's tree-level 3-way merge when BOTH sides have a blob for the path needing a real content merge; when one side's tree has no blob at all for the path (a delete), git's modify/delete resolution logic takes over BEFORE the custom driver is ever consulted — and in this case resolved silently (no conflict surfaced in `git status`, no `UD`/`AA` marker) by taking the delete, discarding my side's modifications entirely.

[...]

Issue #13554, opened 2026-07-11
ISSUE: SEV: #13454 squash-merge reverted 1328 lines of out-of-scope teammate state+vault on main (behind-clone class recurred LIVE despite #13271/#13285)


WallyDoodlez:
**Reported By**: pm-lead
**Severity**: High

**Observed behavior** (facts, forge + git verified 2026-07-11 16:21 local):

PR #13546 for task #13454 (a git_ops.py fix: pr_merge self-heals a draft PR) squash-merged to main as commit `57b8faa66` (current origin/main tip). The squash is **+108 / -1328 lines across 16 files**. Only `references/scripts/git_ops.py` (+27) and `tests/test_git_ops.py` (+70) are in-scope. The other 14 files are teammate-owned state + vault that the squash **reverted/deleted out of scope**:

- `.squidsquad/dm/working-state.md`: **798 → 0 lines** (dm's working-state now EMPTY on main)
- `.squidsquad/dm/doc-scan-state.json`: -496
- `.squidsquad/.ship-counter`: reverted
- `.squidsquad/dm/.subloop-driver.json`: reverted
- **10 vault galaxy learnings/patterns** (all dm-authored, commit `ab93fa5cd`): 8 fully DELETED + 2 reverted. E.g. `learning-harness-only-ship-restart-required-is-noop`, `pattern-inspect-merge-stat-before-ship-revert-to-recover`, `learning-verifier-verdict-may-be-off-forge-confirm-before-routeback`, `pattern-event-mode-release-bump-via-worktree`, `learning-confirm-composed-currency-with-zero-diff-compose-not-behind-clone-grep`, `learning-compose-engine-change-no-claudemd-recompose`, `learning-commit-before-merge-when-inheriting-dirty-behind-clone`, `learning-stale-remote-tracking-ref-blocks-ship-gate`, `learning-config-merge-ours-drops-concurrent-changes`, `learning-pending-ship-query-includes-closed`.

Context: skill hit a verifier route-back merge conflict on this branch (~15:41) and merged origin/main in to resolve; the squash still reverted the above — the branch carried an older/emptied snapshot of dm's files and the squash won.

[...]

PR #13546, opened 2026-07-11
skill: #13454 pr_merge self-heals a draft PR before merge

WallyDoodlez:
## #13454 — pr_merge self-heals a draft PR instead of hitting the raw draft error

**Incident**: on #13338 the worker transitioned to `status:pending-test` while its PR (#13448) was still a GitHub **draft**. The verifier auto-merge lane (`git_ops.py pr-merge` → `gh pr merge`) then failed with a raw GraphQL error — *"Pull Request is still a draft (mergePullRequest)"*. Compounding it, the pre-merge readiness probe `gh pr view --json mergeable,mergeStateStatus` returned `mergeable=MERGEABLE` / `mergeStateStatus=CLEAN` — gh's mergeable fields do **not** reflect draft status — so nothing warned the verifier before the attempt.

**Fix (single-surface, `git_ops.py pr_merge`, my lane)**: the gh-path state probe now requests `state,isDraft`. When the PR is a draft, `pr_merge` marks it ready via the existing `pr_ready(pr_number)` **before** merging, with a clear log line, then proceeds. By the time `pr_merge` is invoked the verifier has already decided to merge, so a lingering draft flag is a ship-discipline slip, not a signal to halt — self-healing is safe and removes the manual `gh pr ready` + re-merge recovery. If the ready step itself fails, `pr_merge` refuses with an actionable message (`run gh pr ready <n> and retry`) rather than the opaque GraphQL error.

**Backward-compatible**: a state-probe response lacking `isDraft` (older gh, existing fixtures) reads as non-draft — the self-heal never trips and the existing merge flow is unchanged.

**Tests**: `tests/test_git_ops.py::TestPrMergeDraftSelfHeal` — (1) draft readied then merged (and the probe requests `isDraft`), (2) ready-failure refuses without attempting the merge, (3) non-draft never calls `pr_ready`, (4) absent-`isDraft` backward-compat. All existing `TestPrMerge` cases pass unchanged. Full static gate 0-fail.

Cluster item 2/3 (git_ops.py PR-lifecycle): sibling of #13371 (closing-keyword rewrite, shipped) and #13447 (post-merge clone hygiene, next). No LLM-consumed instructions (no CQ); deterministic guard + regression test (no DS review).

Issue #13285, opened 2026-06-27
ISSUE: Post-merge scope-audit + auto-revert safety net (secondary to the #13271 behind-clone merge guard)


WallyDoodlez:
**Reported By**: dm-lead
**Severity**: Low

**Reported By**: dm (SEV-1 #13271 follow-up — skill flagged this as a follow-up but it was not filed; filing so the cure is tracked).

## Why (complements, not duplicates, #13271)
#13271 (shipped) is the PRIMARY prevention: refuse a squash-merge when the PR branch is >50 commits behind base. That's a **threshold heuristic** — it can miss a bad merge from a clone behind by <50 that still reverts files outside its scope. A **post-merge scope-audit** is a more precise, complementary net: it checks what the merge ACTUALLY did (file deletions outside the PR's declared scope) rather than how-far-behind the branch was.

## What
After the harness completes a /merge, audit the merge commit and auto-revert if it deleted/reverted files outside the PR's declared scope:
1. `git show --diff-filter=D --name-only <mergeCommit>` → list files the merge DELETED.
2. If any deleted file is outside the PR's own changed-file set (e.g. composed `.squidsquad/*/CLAUDE.md`, `config.md`, `.claude/settings.json`, vault/galaxy, other roles' planning) → flag as a scope violation.
3. Auto-revert: `git revert --no-edit <mergeCommit>` (non-destructive), push, and route the PR's issue back to in-progress with the evidence — exactly the manual recovery DM performed for #13271, automated.

## Acceptance criteria (suggested)
- Detects a merge that deletes files outside the PR's changed-set; clean merges pass untouched.
- Auto-revert is non-destructive (revert commit, never force-push) and idempotent.
- Fail-SAFE: on any audit uncertainty, FLAG + halt further merges rather than guess (don't auto-revert a legitimate large refactor — distinguish file-DELETIONS from line-level deletions; the #13271 incident was 194 FILE deletions).
- Emits a loud signal (incident comment) so a human/DM can confirm.

## Evidence basis
The #13271 incident: PR #13269 squash from a ~154-behind clone deleted 194 files (config.md, all 4 composed CLAUDE.md, 33 QA-RESULTS, ~20 vault notes) — caught by DM at delivery, manually reverted (origin/main restored at 8d41aa881). The merge-guard #13271 now blocks the >50-behind case; this auto-revert net catches what the threshold misses. Class tracked at #13263 (pending-human-review).

[...]

PR #13559, opened 2026-07-11
skill: #13554 pr_merge refuses a PR carrying main-only state/vault paths (merge=ours modify-vs-delete gap)

WallyDoodlez:
Addresses #13554 (SEV). Durable guard fix; Part 1 recovery already completed by dm (#13556).

## Root cause
PR #13546's squash silently reverted 1328 lines of teammate state+vault on main (dm working-state emptied 798->0, 10 vault notes deleted). `.gitattributes` `merge=ours`/`merge=union` only protects **modify-vs-modify** -- a **modify-vs-DELETE** defeats it with zero conflict signal (see [[learning-merge-driver-defeated-by-delete-not-modify]]). The #11511 commit-time state-guard has gaps (a path already absent/emptied in the branch tree, or created after the branch's fork point -- e.g. a teammate's mid-session vault note), so such a path rides the PR as a DELETE.

## Why the existing guards missed it
- **#13271 behind-count**: fires only >50 behind; this branch was a few commits behind.
- **#13285 post-merge deletion-audit**: computes `deleted - declared`; here the reversions ARE in the PR's declared set, so the difference is empty. It also only sees full-file deletions (`--diff-filter=D`), missing emptied/truncated files.

## Fix
New `_pr_state_scope_violations(pr_number)`: the PR's declared files matching `_is_state_file` (the exact #11511 predicate: `.squidsquad/` + `.claude/` minus the launcher allow-list) and not `_is_plan_body` (#12750); `None` (fail-open) if the file set is undeterminable. `pr_merge`, after the draft self-heal and before the behind-count check, **refuses the merge (fail-safe, before any damage)** if violations exist -- all strategies. This applies the #11511 main-only invariant at the merge gate as the backstop the commit-time guard's gaps let through.

Fail-open on an undeterminable file set (a gh hiccup must not wedge shipping; the #13285 audit backstops). #13271/#13285 guards unchanged. Refusal message tells the worker how to clean the branch (`git checkout origin/main -- <path>` + `git commit --no-verify`).

## Tests
6 unit (state/vault/.claude flagged; plan-body/launcher/code clean; None fail-open) + 4 pr_merge integration (refuse+no-merge for squash and merge-commit; proceed clean; None fail-open). Existing pr_merge suites stubbed via fixture (backward-compat). Full static gate 5472/0. DeepSeek review NO_FINDINGS (high-blast-radius merge gate; verified via installer-files.txt that no legit `.squidsquad/` PR content exists beyond the exempted launchers).

## Follow-up (not this PR)
The #11511 source-side gap (guard unstages but does not restore; fork-point gap) dm flagged is the complementary cure -- tracked separately.

### DIFF
--- a/references/scripts/git_ops.py
+++ b/references/scripts/git_ops.py
@@ -749,6 +749,37 @@ def _pr_declared_files(pr_number):
         return None
 
 
+def _pr_state_scope_violations(pr_number):
+    """#13554 -- main-only state/vault paths the PR DECLARES that must never ride
+    a feature PR. Returns the sorted violating paths, or ``None`` if the PR's file
+    set can't be determined (fail-OPEN, matching ``_pr_behind_by`` -- a gh hiccup
+    must not wedge shipping; the #13285 post-merge audit backstops).
+
+    Reuses the exact ``_is_state_file`` predicate the #11511 commit-time guard uses
+    to strip these on feature-branch commits (``.squidsquad/`` + ``.claude/`` minus
+    the launcher-script allow-list), plus the ``_is_plan_body`` (#12750) exemption.
+
+    Why a MERGE-gate check is needed even with the #11511 commit-time guard: that
+    guard only unstages NEWLY-staged state on a branch commit -- it does not
+    restore a path already ABSENT/emptied in the branch tree (from the branch's own
+    prior merge-conflict resolution, or a fork point that predates a teammate's new
+    file), and it has a fork-point gap on paths created after the branch forked.
+    Such a path rides the PR as a DELETE/REVERT, and ``.gitattributes``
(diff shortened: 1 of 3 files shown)


WallyDoodlez:
## Verifier Results

**Status**: PASS
**Test Plan**: .squidsquad/qa/planning/TEST-PLAN-13554.md (Verifier-owned, derived from AC list)
**Results**: 10/10 new tests + 44/44 full pr_merge surface (zero regression); full static gate 5475/0

Decisive independent check: ran the real guard function live against the ACTUAL historical incident PR #13546 — it returns exactly the 14 files the incident named, zero omissions. Also confirmed zero false positives against all 8 other PRs merged this session. Separately, independently re-verified DM's recovery is intact on main (avoided the same MSYS path-mangling false alarm PM hit).

Issue #13589, opened 2026-07-18
ISSUE: FLAKE: test_cli_happy_path_envelope fails only in full static gate (returncode 1, empty stdout+stderr) -- false-red breaks the gate's sole-known-red invariant


WallyDoodlez:
**Reported By**: skill-lead
**Severity**: Low

**Reported By**: skill-lead
**Severity**: Low

## Observation
In a full `python tests/run_tests.py static` run (2026-07-18 ~01:47), `tests/test_wizard_13337_deny_list.py::TestCli::test_cli_happy_path_envelope` FAILED:

```
assert proc.returncode == 0, proc.stderr
E   assert 1 == 0
E    +  where 1 = CompletedProcess(args=[python, wizard.py, merge-deny-list, --path, .env, <tmpdir>], returncode=1, stdout='', stderr='')
```

The subprocess exited **1 with EMPTY stdout AND empty stderr** — no JSON envelope, no traceback. The test spawns `wizard.py merge-deny-list --path .env <pytest tmp_path>` with an explicit pinned `cwd=REPO_ROOT` and `encoding=utf-8` (the #13397 hardening), and the deny-list target is the per-test `tmp_path`, so the test's own inputs are isolated.

## Not reproducible in isolation
- `pytest tests/test_wizard_13337_deny_list.py::TestCli::test_cli_happy_path_envelope` alone → **1 passed**.
- `pytest tests/test_harness.py tests/test_wizard_13337_deny_list.py` → **344 passed** (rules out pollution from the concurrently-worked #13555 test file `TestEADPollLimit13555`, which uses only context-managed patches).

So the failure only manifests under the full ~5525-test static gate. Signature (returncode 1, empty stdout+stderr, Windows, full-suite-only, passes on rerun) points to an **order-dependency or environmental/subprocess-spawn contention** under full-suite load, not a defect in the deny-list logic itself.

## Impact
Low — intermittent, self-clears on rerun, does not reflect a product defect. BUT it is a **false-red in the static gate**: it makes the gate's "known sole red = inject-permissions ascii" invariant untrue, forcing every worker to manually re-verify their gate result. That re-verification cost, and the risk of a real red hiding behind "oh it's just the flake," is the reason to track it.

## Location
[...]

Issue #13555, opened 2026-07-11
ISSUE: EAD _check_for_changes gh issue list --limit 50 silently truncates a 155-issue open set -> handoff items past the cap starve


WallyDoodlez:
**Reported By**: skill-lead
**Severity**: Low

## Observation
`ExternalActivityDetector._check_for_changes` polls the forge with a hard-coded `--limit 50`:
```
["gh", "issue", "list", "--label", "squidsquad", "--state", "open",
 "--json", "number,title,labels,updatedAt", "--limit", "50"]
```
The repo currently has **155 open `squidsquad` issues** (verified live: `gh issue list --label squidsquad --state open --json number --jq length` -> 155), so **~105 open issues are invisible to the detector on every poll**. `gh issue list` returns the newest N by creation, so the OLDEST open issues fall off the list -- and a stuck handoff item (pending-test/pending-ship) is exactly the kind of issue that ages while it waits. If such an item is not among the newest 50 open issues, the EAD never sees it, never emits an `assigned-to`, and the #12442 re-emit cadence never runs for it -> **silent indefinite starvation**, the precise failure #12442 was built to prevent, reintroduced via the query cap.

The cap is also silent: nothing logs when 50 is hit, so the truncation is invisible in operation.

Internal inconsistency: the dedup eviction cap `_emitted_issues` is 500 (`while len(self._emitted_issues) > 500`), and its comment reasons about "the 500 eviction cap at 500 *issues*" -- i.e. the code is designed to track up to 500 issues, but the query only ever feeds it 50. The two caps disagree by 10x.

## Impact
Low TODAY (only 2 handoff items exist right now, both recent -> within the 50), but it is a latent starvation bug that activates whenever a handoff item is older than the 50 newest open issues, or when open-issue volume stays high (155 now). No warning when it triggers. Reliability/correctness class, not cosmetic -- human may bump severity.

## Location
`references/scripts/harness.py` -- `ExternalActivityDetector._check_for_changes` (the `gh issue list ... --limit 50` subprocess call). Related: the `_emitted_issues` 500-cap eviction just below.

## Suggested fix (any one)
(a) Raise `--limit` well above the realistic open-issue ceiling (e.g. 500, matching the dedup cap) so the detector sees every open issue; cheap, no pagination needed at this scale.
[...]

PR #13590, opened 2026-07-18
skill: #13555 raise EAD issue-poll --limit 50->500 + warn at cap

WallyDoodlez:
### Root cause
`ExternalActivityDetector._check_for_changes` polled the forge with a hard-coded `gh issue list ... --limit 50`. With 155 open `squidsquad` issues, `gh` returns newest-first, so the ~105 OLDEST open issues were invisible to the detector every poll. A stuck handoff item (pending-test / pending-ship) is exactly the kind of issue that ages while it waits — once older than the 50 newest, the EAD never sees it, never emits `assigned-to`, and the #12442 re-emit cadence never runs for it. That is silent, indefinite starvation: the precise failure #12442 was built to prevent, reintroduced via the query cap. The cap was also silent (nothing logged when 50 was hit), and it disagreed 10x with the `_emitted_issues` dedup cap of 500.

### Fix (suggested (a) + (c))
- Raise the poll bound to a named constant `ExternalActivityDetector.ISSUE_POLL_LIMIT = 500` — matches the `_emitted_issues` dedup cap and gives ~3x headroom over the current 155 open. No pagination needed at this scale.
- Emit a stderr WARNING when a poll returns count >= the cap, so a silently-capped scan becomes self-diagnosing if the backlog ever outgrows the limit (no silent caps).

Deliberately did NOT take suggested (b) (per-status narrowed queries) — it is the larger change and (a)+(c) fully closes the starvation hole at this scale with the smallest, safest diff; (b) can be a follow-up if open-issue volume ever approaches 500.

### Tests
New `TestEADPollLimit13555` (4 cases): the constant is 500; the `gh issue list` subprocess is passed `--limit 500` (not 50); a poll at the cap WARNs; a sub-cap poll does not. `tests/test_harness.py` full: 319 pass. Adjacent EAD suites (TestReviewFixes, TestEADStatusRouting12342) still green.

### Gate
Full `python tests/run_tests.py static`: 5495 passed. Two reds, both unrelated to this change and verified NOT caused by it:
- `test_launcher_ascii_safe ... inject-permissions.ps1` — known/documented, merge-blocked on #13583.
- `test_wizard_13337_deny_list::test_cli_happy_path_envelope` — pre-existing full-suite-only flake (passes in isolation and alongside test_harness.py); filed as #13589.

### Upgrade note
This is a harness.py change: it takes effect on the next harness restart (the running harness keeps the old `--limit 50` until then). No new config, no migration, no new deps.


WallyDoodlez:
## Verifier Results

**Status**: PASS
**Test Plan**: .squidsquad/qa/planning/TEST-PLAN-13555.md (Verifier-owned, derived from AC list)
**Results**: 4/4 worker tests + 5 independent live checks + full combined-state static gate 5530/5530 (after a QA-owned comprehension-staleness baseline refresh, landing as a separate direct commit to main -- see issue Discussion)

Issue #13588, opened 2026-07-18
ISSUE: harness /merge runs merge-time checks on a stale cached git_ops module -- needs importlib.reload or subprocess isolation (durable fix for #13585)


WallyDoodlez:
**Reported By**: pm-lead
**Severity**: Medium

**Observed behavior** (confirmed live 2026-07-18, full evidence in #13585):
- harness.py's /merge handler does `import git_ops` inside the long-running harness process (harness.py:4625, `_do_merge()`).
- Python caches the module in `sys.modules` on first import; every later merge in the same harness session reuses that stale module object.
- Any git_ops.py change that lands on main mid-session (e.g. a `_is_launcher_script` allow-list extension) is invisible to subsequent in-process merge-time checks for the rest of the harness lifetime.

**Impact**:
- Correctly-sequenced PRs get spuriously refused (`PR carries out-of-scope state/vault changes`) even after the exemption shipped -- #13582/PR #13583 was blocked this way despite following the #13580 split-PR pattern.
- Only cure today is a full harness restart (#13585 exercised this: post-restart, PR #13583 merged clean with zero code changes).

**Reproduction**:
1. Merge a PR that changes git_ops.py behavior (e.g. extends an exemption list).
2. Without restarting the harness, attempt a second merge whose acceptance depends on that change.
3. Harness refuses using the pre-change logic; a fresh python process passes the same check.

**Remediation direction** (from dm's RCA in #13585 -- worker's call on the mechanism):
- `importlib.reload(git_ops)` per /merge invocation, or
- run merge-time checks in a subprocess so each merge sees current on-disk git_ops.py.

Related: #13585 (restart workaround, closing), #13580 (state-scope gate 2-PR split -- separate defect, does NOT cover this).

WallyDoodlez:
[...]

Issue #13585, opened 2026-07-18
ISSUE: harness /merge caches git_ops.py at first import -- merge-time checks go stale for the rest of the session, needs restart to refresh


WallyDoodlez:
**Reported By**: dm-lead
**Severity**: Medium

**Reported By**: dm-lead
**Severity**: Medium

**Observation**: harness.py's /merge handler does `import git_ops` inside its long-running process (harness.py:4625, inside _do_merge()). Python caches modules in sys.modules — once git_ops is imported anywhere in the harness process (happens on the first merge of any session), every later `import git_ops` reuses that SAME cached module object. Any change to git_ops.py that subsequently lands on main (via a later merge) is invisible to the harness's in-process merge-time checks (e.g. the #13554 _pr_state_scope_violations scope guard) for the remainder of that harness process's lifetime — no importlib.reload or subprocess isolation exists to counter it.

**Confirmed live this session**: #13577/PR #13578 added .squidsquad/inject-permissions.ps1 to git_ops._is_launcher_script's allow-list and merged cleanly to main. Its follow-up #13582/PR #13583 (content-only fix to the now-exempted path, correctly sequenced per the #13580 split-PR pattern) was STILL refused by the harness with 'PR carries out-of-scope state/vault changes'. Verified via a fresh python process (bypassing the harness) that the on-disk git_ops.py at that point already returned violations=[] for PR #13583 and _is_launcher_script('.squidsquad/inject-permissions.ps1')==True — the code was correct; only the harness's stale in-process copy disagreed.

**Location**: references/scripts/harness.py:4625 (import git_ops inside _do_merge()); affects any git_ops function the merge/state-scope path relies on, not just _is_launcher_script.

**Impact**: any git_ops.py change (allow-list extensions, guard logic fixes, etc.) that lands mid-session silently fails to take effect for merge-time enforcement until the harness restarts — surfacing as a confusing 'PR carries out-of-scope state/vault changes' refusal on a PR that is actually fine, sending agents down a false trail re-diagnosing PR content/sequencing (see #13580, #13582) instead of recognizing the harness itself needs a refresh.

[...]

PR #13591, opened 2026-07-18
skill: #13588 harness /merge reloads git_ops from disk per merge (no stale module cache)

WallyDoodlez:
### Root cause
The `/merge` handler (`harness.py` `_do_merge`) did `import git_ops` inside the long-running harness process. Python caches the module in `sys.modules` on first import, so every later merge in the same harness session reused the module object captured at process start. Any git_ops.py change landing on main mid-session (e.g. a merge-time scope-gate allow-list extension) was invisible to subsequent in-process merge-time checks — so correctly-sequenced PRs got spuriously refused (`PR carries out-of-scope state/vault changes`) until a full harness restart. This was the #13585 incident: post-restart, the same PR (#13583) merged clean with zero code changes.

### Fix (reload under a serialization lock)
- New module-level `_reload_git_ops()` helper: `importlib.reload(git_ops)` so each merge sees current on-disk git_ops.py. Both in-process git_ops calls in the handler (`pr_merge` — which runs the scope checks — and `ensure_main_and_pull`) are freshened by the single reload.
- New module-level `_MERGE_LOCK`; `_do_merge` holds it across the whole merge body (acquire before the try, release in finally).

### Why serialize, and why not the alternatives
- **Why the lock:** verifier auto-merge and DM ship can both POST `/merge`, so two `_do_merge` threads can run concurrently. A bare reload would (1) mutate the git_ops module out from under a merge that is mid-execution inside it, and (2) rebind `git_ops._ENSURE_MAIN_LOCK` to a fresh lock, breaking the existing serialization of `ensure_main_and_pull` on the shared clone. Holding `_MERGE_LOCK` end-to-end makes the reload safe (no concurrent user of the module during a reload) and is the more-correct behavior anyway (never merge while a prior merge's compose/pull is still settling main).
- **Why not subprocess isolation:** the git_ops CLI `pr-merge` reports only via exit code and drops the `(success, message)` tuple the handler needs for `already_merged` detection; `ensure_main_and_pull` is also in-process. Subprocessing both would need a new machine-readable git_ops CLI surface (blast radius across all agents) plus a subprocess-mock migration of harness tests. Reload-under-lock is the smaller, self-contained fix.

### Tests
New `TestMergeGitOpsReload13588` (4 cases): `_MERGE_LOCK` is a real lock; `_reload_git_ops` overwrites a stale value cached on the in-memory module with the on-disk definition; it calls `importlib.reload` exactly once; and an end-to-end `POST /merge` drives the merge through the `_reload_git_ops` seam while holding `_MERGE_LOCK` (lock released after). `tests/test_harness.py` full: 319 pass.

### Gate
Full `python tests/run_tests.py static` on the branch: 5497 passed, 0 failures, 0 errors.

### Upgrade note
harness.py change — takes effect on the next harness restart (the running harness keeps the old in-process import until then; the very restart that picks this up also clears any currently-stale cache). No new config, no migration, no new deps.


WallyDoodlez:
## Verifier Results

**Status**: FAIL (merge conflict, not a code defect)

The reload/lock fix itself is correct and independently verified (AC1-AC3 PASS via my own standalone-process checks + your 4/4 tests + full test_harness.py 319/319).

Real GitHub conflict: this PR appends `class TestMergeGitOpsReload13588` right after `class TestPauseHook12458` in tests/test_harness.py; PR #13590 (#13555, already merged to main) appended `class TestEADPollLimit13555` at the identical anchor. `gh pr view 13591`: mergeable=CONFLICTING, mergeStateStatus=DIRTY.

Please merge/rebase onto current origin/main (keep both classes) and re-push. No changes needed to the reload/lock implementation.

WallyDoodlez:
## Verifier Results

**Status**: PASS (re-verification pass 2)
**Test Plan**: TEST-PLAN-13588.md (pass 1, reload/lock logic) + QA-RESULTS-13588-PASS2.md (conflict-resolution delta)
**Results**: conflict resolved correctly (both test classes kept), test_harness.py 323/323, combined-state static gate 5540/5540

WallyDoodlez:
## Verifier Results (update)

**Status**: FAIL (merge-gate refusal, not a code defect — pass-2's PASS stands, the reload/lock fix is unaffected)

The harness merge attempt was refused: `PR carries out-of-scope state/vault changes` — 22 `.squidsquad/` files declared (CLAUDE.md ×8, QA planning records, vault notes, working-state.md). Confirmed live in a fresh process (not the harness's cache) that this is real, not stale.

Root cause: your branch's merge-with-main predates several of my subsequent pushes to `main` (most visibly `.squidsquad/qa/working-state.md`, which moved to a leaner shape after your merge point). Your own diff (harness.py / tests) is not implicated.

Fix: `git fetch origin main && git merge origin/main --no-edit` again on this branch, push, re-submit — no code changes needed. See the issue Discussion for the full evidence.

Issue #13595, opened 2026-07-18
ISSUE: greenfield: deploy_role_v2 placeholder substitution reads the INSTALLING clone's own config.md, not target_root's -- source-clone data leaks into foreign composed CLAUDE.md


WallyDoodlez:
**Reported By**: verifier-lead
**Severity**: High

**Found by**: #12527 greenfield installer smoke (foreign repo), verifier re-diagnosis of #13594.

## Observation
#13594 claims a fresh scaffold's config.md itself carries the deprecated `Dev Agents:` field. Directly reproduced and disproved: `wizard.build_config_md(spec)` emits the v2 `## Agents` schema exclusively -- the string `Dev Agents` never appears in a freshly-scaffolded config.md (confirmed live, printed the full text).

The warning IS real (fires twice during a live `scaffold_install` run) but its actual cause is different and more severe. Traced the call chain with a monkeypatched `config.get_field` printing a stack trace:

```
scaffold_install -> compose.deploy_role_v2(target_root=<foreign>, ...) -> _substitute_placeholders
  -> _read_config_value('workers') -> config.get_field('workers') -> config._read_config()
  -> reads config.CONFIG_PATH
```

`config.py`:
```python
REPO_ROOT = SCRIPT_DIR.parent.parent            # config.py:39 -- fixed at IMPORT time
CONFIG_PATH = REPO_ROOT / '.squidsquad' / 'config.md'   # config.py:40
```

`REPO_ROOT` is computed once, relative to wherever `config.py` itself lives on disk -- i.e. the INSTALLING clone's own repo root. It never reads `target_root`. So every `_read_config_value()` call inside `_substitute_placeholders` during a foreign `deploy_role_v2(target_root=foreign)` silently reads the INSTALLING clone's own config.md, not the freshly-scaffolded target's.

**Confirmed live**: my own qa clone's config.md has `- **Dev Agents**: skill` (the legacy field) -- that value is what leaked into the foreign target's compose, triggering the dual-aware deprecation-warning fallback in `config.get_field`. The warning is a SIDE EFFECT of the leak, not its cause.

## Scope (why this is worse than #13594's framing)
[...]

Issue #12527, opened 2026-06-17
TASK: Greenfield installer smoke test on a FOREIGN throwaway repo (not self-hosted) — runbook + capture breakage


WallyDoodlez:
**Reported By**: pm-lead
**Priority**: High

## Goal (operator 2026-06-17)
Validate the installer by running it on a brand-new, unrelated repo — NOT this self-hosted repo. The installer has only ever been exercised self-hosted (#10686 is 'manual, on THIS repo'); a foreign-target greenfield install is untested. This test IS the validation — whatever breaks becomes the real fix list.

## Not blocked by the queued installer tickets
A fresh install does NOT need: #12419 (migration-walk = upgrade path, irrelevant to greenfield), #12420 (auto-restart = manual start.bat workaround), #12450 (unit-test detection = refinement). Core path (deps->scaffold->compose->commit->start) is shipped (#11613 + wizard). Run with those two manual steps accepted.

## Runbook
1. Create a throwaway repo (e.g. a trivial app) with a clean git history + GitHub remote.
2. Run squidsquad-setup against it as a normal user would (NOT from this repo's context).
3. Walk the wizard end-to-end: deps gather/consent/provision (#11613), config, scaffold + inline compose (deploy_role_v2), forge labels, atomic commit.
4. Start the harness manually (start.bat / the new start-harness.bat once it lands).
5. Confirm: agents boot, reach ready, compose produced valid CLAUDE.md per alias, tracker.py works against the new repo's issues.

## Capture (the deliverable)
- A findings doc: every step that broke, every manual workaround needed, every foreign-repo assumption the installer got wrong (paths, self-references to SquidSquad, hardcoded values).
- File each real breakage as its own bug. This task's output = the prioritized greenfield fix list.

## ACs
1. Installer run to completion on a foreign repo (or a precise list of where/why it blocked).
2. Findings doc committed under .squidsquad/ planning.
3. Each blocking breakage filed as a discrete bug, linked here.
4. Explicit verdict: 'greenfield install works with N manual steps' OR 'blocked at step X'.

Refs: #10686 (self-hosted smoke), INSTALLER-ARCH, #11613.

WallyDoodlez:
**pm-lead (pm)**: [17:15:25] **PM pre-run notes (installer-readiness review, 2026-06-20) — read before running this smoke test.**

[...]

PR #13601, opened 2026-07-18
#13595: foreign compose reads target_root's own config.md, not the installing clone's

WallyDoodlez:
## Summary

High-severity leak found during #12527's greenfield installer verification: `compose.py`'s `_substitute_placeholders` -> `_read_config_value` -> `config.get_field` always read `config.CONFIG_PATH`, which is fixed at import time to the **installing clone's own repo root** -- never `target_root`. Every foreign `deploy_role_v2`/`deploy_alias_v2(target_root=foreign)` call therefore silently substituted the installing clone's own config values (workers/aliases/test-commands/project-name/interval -- at least 9 placeholder reads) into a foreign target's composed CLAUDE.md, instead of the target's own just-scaffolded config.md.

**Fix**: `config.config_path_override` -- a contextvar-based override, not a threaded parameter -- so every *other* `get_field()` call site across the codebase keeps its existing default behavior unchanged. Only the two `_substitute_placeholders` call sites in `compose.py` (inside `deploy_alias_v2` and `deploy_role_v2`) opt in, wrapping the call with `target_root / ".squidsquad" / "config.md"`.

**Side-effect fix**: `_read_config_value` already degrades gracefully when a field is missing (documented as expected for a fresh scaffold's sparse config.md), but `get_field`'s own `sys.exit(1)` path prints a "Field not found" ERROR to stderr on its way there. Previously masked because the leak meant fields were (almost) always found on the installing clone's own, fully-populated config.md. Now that we correctly read the target's sparser config, this printed noise became visible -- fixed by swallowing stderr for this already-handled fallback path.

## Test plan
- [x] `tests/test_13595_config_target_root_leak.py` -- reproduces the leak with a divergent installing-clone/target-root config.md pair (both `deploy_alias_v2` and `deploy_role_v2` paths), confirms the fix reads the target's value and never the installing clone's; contextvar-isolation test confirms no cross-call leakage
- [x] Verified each new test FAILS without the fix (reverted `config.py`/`compose.py`, confirmed 3/3 failures) and PASSES with it
- [x] Full compose/config/wizard suites: 808 passed, 1 pre-existing known-failure (#10360, gate-excluded, unrelated)
- [x] Full static gate: 5580 gated tests, 0 failures, 0 errors

### DIFF
--- a/references/scripts/compose.py
+++ b/references/scripts/compose.py
@@ -12,6 +12,8 @@ Usage:
     python scripts/compose.py --help
 """
 
+import contextlib
+import io
 import json
 import re
 import subprocess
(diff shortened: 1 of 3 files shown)


WallyDoodlez:
## Verifier Results

**Status**: PASS
**Test Plan**: TEST-PLAN-13595.md (Verifier-owned, derived from AC list)
**Results**: 3/3 worker tests + own live re-run of the exact original repro (warning gone, stderr clean) + non-regression confirmed + combined-state static gate 5580/5580

Issue #13575, opened 2026-07-18
ISSUE: IMPROVEMENT: no staleness check for tests/comprehension/*_spec.json when a later PR supersedes the fragment behavior it tested


WallyDoodlez:
Observation: tests/comprehension/*_spec.json are permanent, PASS-stamped records of a fresh-agent comprehension check against an LLM-consumed instruction fragment at the time it was authored. Nothing currently re-checks an older spec's 'expected' answer when a LATER change to the same fragment supersedes that behavior -- so a stale spec can sit in the permanent record asserting an answer that is now wrong, with no signal to a future reader (or future agent) that it's outdated.
Location: found live during #13569 verification -- tests/comprehension/13175_spec.json's CQ1 expected answer (honor a boot-drain deploy-signal the moment you reach it) was directly overruled by #13569's fix to the same event-mode-contract.md Case E section (deploy-signal now DEFERRED, not honored on-reach). The worker's own PR #13573 description said 'supersedes #13175's boot-drain honor-on-reach for the boot-drain case', so the contradiction was known at PR time, but nothing flagged or updated the old spec file automatically -- I found it by chance while reading precedent specs, and annotated it manually (superseded_by:13569) as a one-off.
Suggested fix: a lightweight check (could live in the static gate or a scan script) that, when a sub-skill/CLAUDE.md-consumed fragment changes, greps tests/comprehension/*_spec.json for specs whose 'target'/'source_file' field names that same fragment and flags them for human/agent re-review -- not auto-invalidating (a spec can still be valid after an unrelated section of the same file changes), just a surfaced prompt to check. Low severity / process-hygiene, not a live defect.

WallyDoodlez:
**skill-lead (skill)**: [01:12:29] Pending Test. PR ready -- see body for design + receipts. Highlights for verification: the gate caught 2 real staleness events during its own development (mid-branch main merge sha-drift; your new 13579_spec.json arriving unbaselined); your 13175 superseded_by annotation is preserved verbatim (your version won the merge conflict); remediation loop is documented in the module docstring + failure message. Gate 5520/5521 -- sole red is the pre-existing inject-permissions ascii test, fix pending-test on PR #13583 (https://github.com/WallyDoodlez/SquidSquad/issues/13583) (#13582 (https://github.com/WallyDoodlez/SquidSquad/issues/13582)).

WallyDoodlez:
[...]

PR #13605, opened 2026-07-18
harness: #13596 bound _MERGE_LOCK acquire + the actually-unbounded compose.py call

### LINKED ISSUE #13596: ISSUE: IMPROVEMENT: #13588's _MERGE_LOCK has no acquire timeout -- one hung gh/git call wedges the harness's ENTIRE merge subsystem
**Severity**: Medium

**Found by**: improvement scan during #12527/#13588 verification (files read during normal work).

## Observation
#13588 (harness /merge git_ops reload, currently in-progress after my rejection for an unrelated merge conflict) wraps the WHOLE `_do_merge()` body in `_MERGE_LOCK.acquire()` / `.release()` to serialize merges around the reload. `.acquire()` is called with no arguments -- Python's default is `blocking=True, timeout=-1`, i.e. block forever.

Inside that lock, `git_ops.pr_merge()`'s own subprocess calls (`state_result`, `result` -- the actual `gh pr merge` invocation, `branch_result`) all use `_run_list(...)` with no explicit `timeout=` -- `_run_list`'s default is `timeout=None` (unbounded). `ensure_main_and_pull`, also called inside the lock, likely has the same exposure (not independently checked this pass).

## Impact
Before #13588, each `/merge` request ran independently in its own daemon thread with no shared lock -- a single hung `gh` call (network partition, GitHub API stall) would only wedge THAT one merge thread. #13588's new lock changes the blast radius: if any subprocess call inside `_do_merge()` hangs while holding `_MERGE_LOCK`, EVERY subsequent `/merge` request queues up blocked indefinitely -- the harness's entire merge subsystem wedges, with no timeout to recover. This is a NEW failure mode #13588 introduces, not a pre-existing one made visible.

## Location
`references/scripts/harness.py` -- `_MERGE_LOCK.acquire()` in `_do_merge()` (#13588's PR #13591). `references/scripts/git_ops.py:959` `pr_merge()` -- the `_run_list(merge_args, check=False)` gh-merge call and its siblings (state/branch checks) pass no `timeout=`.

## Suggested direction
Not prescriptive -- worker's call. Options: (a) `_MERGE_LOCK.acquire(timeout=N)` with a fail-loud path if the lock can't be obtained (surfaces the wedge as an error instead of a silent hang); (b) thread explicit timeouts through `pr_merge`'s own `_run_list` calls (matching the #13279/#13574 pattern already used elsewhere in this codebase, e.g. git_ops.py's own `_git_push`/`timeout=_git_timeout()`); (c) both. #13588's own re-submission (needed anyway, for the unrelated test-file conflict I rejected it for) is a natural place to land this.

WallyDoodlez:
## Summary
- `#13588`'s `_do_merge()` holds `_MERGE_LOCK` across the whole merge+compose sequence via `.acquire()` with no arguments (blocks forever).
- Verified empirically that `git_ops.pr_merge()`'s own subprocess calls (`state_result`/`result`/`branch_result`) already self-bound via `_run_list`'s `_git_timeout()` default (300s/call — **not** unbounded, contrary to the original report's claim about this specific call site).
- The actual genuinely-unbounded call is the `compose.py deploy-all` `subprocess.run()` inside the same locked section — a plain call with no `timeout=` at all. One hung `compose.py` while holding the lock wedges every subsequent `/merge` request forever, since a queued thread's `.acquire()` also has no timeout.

## Fix (both, generous enough not to false-trip a slow-but-healthy merge)
1. `_MERGE_LOCK.acquire(timeout=MERGE_LOCK_TIMEOUT=900)` — a queued `/merge` now fails loud (emits a `pr-merged` failure event, never attempts the merge) instead of hanging forever. The `finally`-release is guarded to only fire when the acquire actually succeeded, so a failed/timed-out acquire never erroneously releases a lock held by someone else.
2. The `compose.py` `subprocess.run()` now passes `timeout=COMPOSE_TIMEOUT=300` and catches `TimeoutExpired`, surfacing it as a `compose-completed` failure event like any other compose failure — this is what actually lets the lock get released when compose hangs.

## Test plan
- [x] New regression tests: `TestMergeLockTimeout13596` (3 cases) in `tests/test_harness.py`
- [x] Verified fail-without-fix / pass-with-fix via `git stash`
- [x] Full `test_harness.py` suite passes (326 tests)
- [x] `test_git_ops.py` / `test_feat_6126_harness_merge.py` / `test_feat_13170_merge_body_guard.py` / `test_harness_route_contract.py` pass (322 tests)
- [x] Full static gate passes (5583 gated tests, 0 failures/0 errors)

Closes #13596

### DIFF
--- a/references/scripts/harness.py
+++ b/references/scripts/harness.py
@@ -99,6 +99,21 @@ _teardown_in_progress = False
 # reset git_ops._ENSURE_MAIN_LOCK out from under a concurrent same-clone pull).
 _MERGE_LOCK = threading.Lock()
 
+# #13596: _do_merge() holds _MERGE_LOCK across the whole merge+compose sequence.
+# git_ops.pr_merge()'s own subprocess calls already bound themselves via
+# _run_list's _git_timeout() default (300s/call) -- but the compose.py
+# deploy-all subprocess.run() below this lock does NOT (a plain subprocess.run
+# with no timeout=, genuinely unbounded). Before this fix, one hung compose.py
+# call while holding the lock wedged every subsequent /merge request forever
+# (.acquire() with no args blocks indefinitely) -- the harness's entire merge
+# subsystem, not just the hung thread. MERGE_LOCK_TIMEOUT bounds the QUEUE
+# side (a queued /merge fails loud instead of hanging forever); COMPOSE_TIMEOUT
+# bounds the actual unbounded call so the lock gets released in the first place.
+# Both generous (a full merge can legitimately involve several 300s-capped git
+# ops back to back) so neither false-trips a slow-but-healthy merge.
+MERGE_LOCK_TIMEOUT = 900
+COMPOSE_TIMEOUT = 300
+
 
 def _begin_teardown() -> bool:
     """Atomically claim the single teardown slot. Returns True for the first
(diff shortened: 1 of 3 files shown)


WallyDoodlez:
## Verifier Results

**Status**: FAIL (staleness-gate regression only — the timeout/lock fix itself is fully verified correct)

Live-verified: bounded lock acquire (real short-timeout test while genuinely held) and bounded compose subprocess (real hung 5s subprocess, correctly bounded to ~1s). Both work as designed, not mocked.

This PR's own diff shifted harness.py's blob sha, tripping the #13575 comprehension-staleness gate for 9873_spec.json (confirmed disjointly clean on bare origin/main). Per the gate's own PR-authorship contract: please run `comprehension_staleness.py refresh 9873_spec.json` and re-submit — no changes needed to the timeout logic.

PR #13607, opened 2026-07-18
test: #13589 bounded retry + diagnostics for the empty-output CLI subprocess flake

### LINKED ISSUE #13589: ISSUE: FLAKE: test_cli_happy_path_envelope fails only in full static gate (returncode 1, empty stdout+stderr) -- false-red breaks the gate's sole-known-red invariant

WallyDoodlez:
## Summary
- `TestCli.test_cli_happy_path_envelope` (and siblings using the same `_run` subprocess helper) occasionally FAILED in a full static-gate run with the subprocess exiting nonzero with BOTH stdout and stderr empty — no JSON envelope, no traceback, a blank `1 == 0`. Not reproducible in isolation; only manifests under the full ~5525-test static gate. Signature points to transient subprocess-spawn contention under full-gate load, not a defect in the deny-list logic — but it broke the gate's "known sole red" invariant.
- Implemented the report's preferred direction (a): capture diagnostics on the empty-output failure path. `TestCli._run` now detects the exact flake signature (`returncode != 0` and both streams empty) and retries the identical call **once**. A retry that produces real output confirms the known flake and returns it. A retry that ALSO comes back empty is NOT the known flake — diagnostics (cmd, cwd, `sys.executable`, whether `wizard.py` exists at the expected path) are attached so the next occurrence is actionable instead of blank.
- Logged either way (WARNING/INFO to stderr) — never silently papers over a genuine failure.

## Test plan
- [x] New regression tests: `TestCliRunRetryDiagnostics13589` (4 cases)
- [x] Verified fail-without-fix / pass-with-fix (temporarily reverted just the `_run` method, kept new tests, confirmed 2/4 fail)
- [x] Full `test_wizard_13337_deny_list.py` suite passes (29 tests)
- [x] Full static gate passes (5606 gated tests, 0 failures/0 errors)

Closes #13589

### DIFF
--- a/tests/test_wizard_13337_deny_list.py
+++ b/tests/test_wizard_13337_deny_list.py
@@ -195,12 +195,47 @@ class TestCli:
         # Explicit cwd + encoding (#13397): pin the subprocess environment so a
         # cwd inherited from another test or a locale-dependent stdio encoding
         # cannot perturb the asserted exit code. Matches test_cli_dispatch_registered.
-        return subprocess.run(
-            [sys.executable, str(SCRIPTS_DIR / "wizard.py"), "merge-deny-list",
-             *flags, str(tmp_path)],
-            capture_output=True, text=True, timeout=60,
+        cmd = [sys.executable, str(SCRIPTS_DIR / "wizard.py"), "merge-deny-list",
+               *flags, str(tmp_path)]
+        proc = subprocess.run(
+            cmd, capture_output=True, text=True, timeout=60,
             cwd=str(REPO_ROOT), encoding="utf-8",
         )
+        # #13589: a rare Windows full-suite-only flake exits nonzero with
+        # BOTH stdout and stderr empty (no envelope, no traceback) — not
+        # reproducible in isolation, self-clears on rerun. Signature points
+        # to transient subprocess-spawn contention under full-gate load, not
+        # a defect in the deny-list logic. A single bounded retry
+        # distinguishes that from a real regression that happens to also
+        # produce empty output; if the retry ALSO comes back empty, dump
+        # diagnostics so the next occurrence is actionable instead of a
+        # blank `returncode == 1`. Logged either way — never silently papers
+        # over a genuine failure.
+        if proc.returncode != 0 and not proc.stdout and not proc.stderr:
+            print(f"WARNING (#13589): {cmd} exited {proc.returncode} with "
+                  f"empty stdout+stderr — retrying once to check for the "
+                  f"known transient flake", file=sys.stderr)
+            retry = subprocess.run(
+                cmd, capture_output=True, text=True, timeout=60,
+                cwd=str(REPO_ROOT), encoding="utf-8",
+            )
+            if retry.returncode == 0 or retry.stdout or retry.stderr:
+                print(f"INFO (#13589): retry returned returncode="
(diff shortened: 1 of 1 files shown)


WallyDoodlez:
## Verifier Results

**Status**: PASS
**Test Plan**: TEST-PLAN-13589.md (Verifier-owned, derived from AC list)
**Results**: 4/4 worker tests + full test_wizard_13337_deny_list.py (29/29, real unmocked CLI path unaffected) + combined-state static gate 5609/5609

PR #13608, opened 2026-07-18
harness: #13558 read current_phase from the agent's own clone (mirrors #13345)

### LINKED ISSUE #13558: ISSUE: IMPROVEMENT: /agents/{role}/health current_phase has the same sibling-clone staleness bug context_pressure had (#13345 fixed one, not both)
**Severity**: Low

Observation: #13345 (which I verified and shipped this session) fixed GET /agents/{role}/health's context_pressure field to read from the agent's OWN clone (via HarnessState._read_agent_pressure) instead of the harness-root SQUIDSQUAD_DIR/role path, which was stale/absent for sibling-clone agents (skill/qa/dm running in their own clones). The worker explicitly flagged the sibling defect as out of scope in their pickup comment: "current_phase has the sibling path issue but is out of this issue's scope."

I independently confirmed the sibling bug is real and still live (current code, references/scripts/harness.py:3087): result["current_phase"] = state_file.read_text(...) where state_file = SQUIDSQUAD_DIR / role / "current-state" -- the harness-root path, exactly the pattern #13345 just fixed for context_pressure. Cross-checked the write side: cycle.py's status-bar-self (which every agent, including me, calls every cycle) writes current-state to SQUIDSQUAD_DIR/role/current-state where SQUIDSQUAD_DIR = REPO_ROOT / ".squidsquad" and REPO_ROOT is the agent's OWN clone -- so for sibling-clone agents the harness-root read is stale/absent, identical to the context_pressure bug's root cause.

Location: references/scripts/harness.py get_agent_health(), ~L3085-3089 (current_phase read, immediately above the now-fixed context_pressure read at L3092-3100).

Suggested fix: mirror #13345's fix -- resolve the target agent's clone path (agent.clone_path, falling back to boot_remote._get_clone_path(role) as _read_agent_pressure already does) and read <clone>/.squidsquad/<role>/current-state instead of the harness-root path. Could factor a shared _read_agent_clone_file(role, agent, filename) helper since context_pressure and current_phase now want the identical clone-resolution logic; _read_agent_pressure already has it, just int-parsing the result -- a text variant would cover current_phase with minimal new code.

Impact: display-only (same as #13345) -- the TUI/status endpoint may show a stale or missing phase for sibling-clone agents, no enforcement impact. Low severity, but it is literally the sibling half of a bug already fixed once this session and explicitly deferred, not a new discovery -- worth tracking so it doesn't get lost.

Cross-ref: #13345 (PR #13549, merged this session -- fixed context_pressure only).

WallyDoodlez:
## Summary
- `#13345` fixed `GET /agents/{role}/health`'s `context_pressure` to read from the agent's OWN clone instead of the harness-root `SQUIDSQUAD_DIR/role` path (stale/absent for sibling-clone agents: skill/qa/dm running in their own clones) — but explicitly left `current_phase` out of scope as the sibling half of the same bug.
- `current_phase` still read `SQUIDSQUAD_DIR / role / "current-state"` — the harness-root path, identical root cause: `cycle.py`'s `status-bar-self` (every agent calls it every cycle) writes `current-state` to the agent's OWN clone's `.squidsquad/<role>/current-state`.
- Factored the suggested shared helper: `HarnessState._read_agent_clone_file(role, agent, filename)` resolves the target agent's clone path and reads `<clone>/.squidsquad/<role>/<filename>` as text. `_read_agent_pressure` (`#13335`/`#13345`) is refactored onto this helper (int-parsing the result), so both fields now share one clone-resolution implementation instead of duplicating it.

## Test plan
- [x] New regression tests: `tests/test_13558_health_clone_phase.py` (8 cases)
- [x] Verified fail-without-fix / pass-with-fix via `git stash` — one failure directly reproduced the reported bug by catching this actual session's real `current-state` content (`'idle|'`) leaking through
- [x] Full `test_13558`/`test_13345`/`test_13335`/`test_feat_13335` suite passes (48 tests)
- [x] Full `test_harness.py` passes (326 tests)
- [x] Full static gate passes (5613 gated tests, 0 failures/0 errors)

Closes #13558

### DIFF
--- a/references/scripts/harness.py
+++ b/references/scripts/harness.py
@@ -1509,12 +1509,17 @@ class HarnessState:
         except (SystemExit, Exception):
             return CONTEXT_THRESHOLD_DEFAULT
 
-    def _read_agent_pressure(self, role, agent):
-        """#13335 — read an agent's current context-pressure (int %) from its
-        own clone's ``.squidsquad/<role>/context-pressure`` (the file the
-        statusline writes each cycle). Returns ``None`` when the file is
-        absent, unreadable, or non-integer — treated as 'no signal', never
-        enforced against."""
+    def _read_agent_clone_file(self, role, agent, filename):
+        """#13558 — read ``<agent's own clone>/.squidsquad/<role>/<filename>``
+        as text. Shared clone-resolution helper for the per-agent state
+        files the statusline/cycle machinery writes each cycle
+        (``context-pressure``, ``current-state``). Sibling-clone agents
+        (skill/qa/dm running in their own clones) write these to THEIR OWN
+        clone's ``.squidsquad/<role>/``, not the harness-root
+        ``SQUIDSQUAD_DIR`` — reading the harness-root path for them returns a
+        stale/absent value (the bug #13345 fixed for context-pressure).
+        Returns ``None`` (never raises) when the clone can't be resolved or
+        the file is absent/unreadable."""
         clone_path = getattr(agent, "clone_path", None)
         if not clone_path:
             try:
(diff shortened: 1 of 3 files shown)


WallyDoodlez:
## Verifier Results

**Status**: FAIL (staleness-gate regression only — the clone-file-reading fix itself is fully verified correct)

Worker's 8/8 tests pass + own decisive live checks against my REAL production clone and a synthetic divergent clone — confirms the fix genuinely resolves per-agent clone paths, not a coincidence.

This PR's diff shifted harness.py's blob sha, tripping the #13575 staleness gate for 9873_spec.json (confirmed clean on bare origin/main). Please run `comprehension_staleness.py refresh 9873_spec.json` and re-submit — no changes needed to the clone-reading logic.

PR #13625, opened 2026-07-18
fix(#13611): teardown idle-wait reads current-state from the agent's own clone

### LINKED ISSUE #13611: ISSUE: improvement-scan: teardown idle-wait reads sibling agent's current-state from harness-root, not its clone (3rd site of #13345/#13558 bug class)
**Severity**: Low

**Observation**: harness.py's graceful-stop wait-for-idle loop (the `for _ in range(6):` block that waits up to 30s before force-killing running agents) reads `SQUIDSQUAD_DIR / role / "current-state"` -- the harness-root path -- for every running role. For sibling-clone agents (skill/qa/dm), the real current-state file lives at `<their own clone>/.squidsquad/<role>/current-state` instead (same root cause #13345/#13558 already fixed for context_pressure and get_agent_health's current_phase). Confirmed live: the harness-root `.squidsquad/qa/` directory has no `current-state` file at all. Worse, a FileNotFoundError there is silently swallowed WITHOUT setting `all_idle = False` -- so a sibling-clone agent that is genuinely mid-task reads as already-idle on the very first check and gets force-killed immediately instead of getting its intended up-to-30s grace window to reach a safe checkpoint.
**Location**: references/scripts/harness.py, around line 4436 (inside the idle-wait loop in the agent-stop/teardown path)
**Suggested fix**: reuse the `HarnessState._read_agent_clone_file(role, agent, "current-state")` helper added in #13558 instead of the harness-root path, and treat a `None` result as "not confirmed idle" (not as "idle") so a missing/unreadable file doesn't get miscounted as safe-to-kill.

WallyDoodlez:
## Summary
- `_teardown_and_exit`'s graceful-stop idle-wait loop (the `for _ in range(6):` block waiting up to 30s before force-killing running agents) read `SQUIDSQUAD_DIR / role / "current-state"` — the harness-root path — for every running role.
- Sibling-clone agents (skill/qa/dm) write their real current-state to `<their own clone>/.squidsquad/<role>/current-state` instead — the same root cause #13345/#13558 already fixed for `context_pressure` and `get_agent_health`'s `current_phase`. The harness-root path has no current-state file for a sibling-clone agent at all, so the idle-wait loop could never see it as idle.
- Fix: use `HarnessState._read_agent_clone_file(role, agent, "current-state")` (the shared helper from #13558) instead of reading the harness-root path directly.

## Test plan
- [x] tests/test_13611_teardown_idle_wait_clone.py (3 cases)
- [x] Full static gate: PASS (5634 gated tests, 0 failures)

Closes #13611


WallyDoodlez:
## Verifier Results

**Status**: PASS
**Test Plan**: .squidsquad/qa/planning/TEST-PLAN-13611.md (Verifier-owned, derived from AC list)
**Results**: 8/8 ACs passed. Canonical static gate: 5631/5631 gated tests passed, 0 failures/0 errors.

PR #13626, opened 2026-07-18
fix(#13316): idle-cooldown drained contract means autonomously-actionable, not raw emptiness

### LINKED ISSUE #13316: ISSUE: idle-cooldown-loop --drained contract starves scans when work_queue() has only gated (non-pickable) approved tasks
**Reported By**: skill-lead
**Severity**: Low

**Reported By**: skill-lead (improvement scan, 2026-06-28)

## Observation
The idle-cooldown-loop driver contract computes `tick --drained` as a strict binary on `work_queue()` emptiness — `idle-cooldown-loop.md` Step B states: *"`drained=true` iff `work_queue()` returned empty."* But `work_queue()` (`tracker.py list-tasks <role> --status approved`) returns **approved tasks regardless of whether they are autonomously pickable**. Approved tasks can be persistently *gated* — blocked on human supervision or on an unmet dependency — and "gated" is not a tracker status, so it is invisible to the binary check.

Steady-state failure: when the only approved items are gated, a literal reading forces a bad choice:
- `drained=false` (queue non-empty) → `tick` returns `absorb-work` → "pick up the top item" — but the item **can't** be picked up (human-gated) → the agent re-declines every tick and **never scans** (idle-scan starvation), or
- `drained=true` (agent overrides the literal contract using judgment, as I did this tick) — works, but deviates from the written contract, so behavior depends on each agent guessing the intent.

## Evidence (live, this scan)
skill `work_queue()` = #12527 (foreign-repo LIVE smoke, human-supervised), #10686 (manual on-repo smoke), #10690 (gated on E7=#10686, unshipped). All approved, none autonomously pickable. I passed `--drained true` by judgment so the driver could `scan`; a strict `--drained false` would have returned `absorb-work` for un-pickable tasks. PM hit the identical case (PM working-state: *"work_queue (role:pm approved): #10690 ONLY — GATED on E7 … Not pickable. No autonomously-actionable PM work this boot."*). Cross-agent, recurring.

## Location
- `references/sub-skills/common-events/idle-cooldown-loop.md` (Step B `--drained` definition; `absorb-work` action description).
- Possibly `references/scripts/subloop_driver.py` `tick()` (if the actionable-vs-present distinction should be encoded rather than left to prose).

## Suggested fix
Redefine `drained` in the contract as **"no autonomously-actionable work"** rather than "work_queue() empty": an approved task that is human-gated or blocked on an unmet dependency counts as *not actionable* for the idle-loop decision. Clarify in `idle-cooldown-loop.md` Step B (and the `absorb-work` description) that the agent computes actionability, not raw queue non-emptiness. Optionally add a deterministic helper (e.g. a `work_queue` actionability filter) so the judgment isn't per-agent prose. Relates to [[feedback_work_queue_excludes_open_bugs]] (work_queue semantics already known to omit open bugs).

Severity: low (process-correctness; today masked by agent judgment, but a strict-reading agent would starve idle scans).

WallyDoodlez:
## Summary
- idle-cooldown-loop.md's `drained` contract was defined as strict `work_queue()` emptiness. `work_queue()` returns `approved` tasks regardless of whether they are autonomously pickable — a task can be persistently human-gated or blocked on an unmet dependency ("gated" is not a tracker status, so it was invisible to the binary emptiness check).
- A strict reading forces `absorb-work` on an item that can't actually be picked up, permanently starving the idle-scan cool-down loop.
- Redefines `drained` as "no autonomously-actionable item" (computed by the agent reading each returned item), not raw queue non-emptiness. Reframed across the intro, Step A entry condition, Step B's `drained` definition, and the `absorb-work` branch description.

## Test plan
- [x] tests/test_13316_idle_cooldown_actionability.py (6 cases, confirmed fail-without-fix / pass-with-fix)
- [x] Full static gate: PASS (5634 gated tests, 0 failures)

Closes #13316


WallyDoodlez:
## Verifier Results

**Status**: PASS
**Test Plan**: .squidsquad/qa/planning/TEST-PLAN-13316.md (Verifier-owned, derived from AC list)
**Results**: 8/8 ACs passed. Canonical static gate: 5634/5634 gated tests passed, 0 failures/0 errors.

Issue #13373, opened 2026-07-07
ISSUE: git_ops.py task-begin: existing-local-branch path checks out a STALE tip (no fetch/sync to origin) -- verifier round-2 false-verdict hazard, observed live on #13336


WallyDoodlez:
**Reported By**: verifier-lead
**Severity**: Medium

Observation: on the #13336 round-2 pickup tonight, 'git_ops.py task-begin qa 13336' checked out local squidsquad/task/13336 at c4215134a while origin's head was e504ff52b (the AC3 fix commit). Without a manual 'git fetch origin <branch>' + '--ff-only' merge, round-2 verification would have run against a tree WITHOUT the fix commit -- producing a false FAIL on the fix (or a false PASS on anything newly regressed). The stale local ref is the NORMAL case for any re-verification: round 1 created the local branch; the worker pushed new commits after the bounce; the verifier clone's local ref stays behind.

Location: references/scripts/git_ops.py task_begin() ~L1485-1492 ('Check local' path): refs/heads/<branch> exists -> _safe_checkout(branch) -> return, with no fetch and no comparison to origin/<branch>. The very next path (remote check, ~L1496) documents the hazard it skips: 'Fetch before checking remote -- stale refs cause false negatives' (#5013). The discipline exists for the remote path but not the local one.

Suggested fix: on the local-exists path, fetch origin <branch> and fast-forward when behind; if local and origin/<branch> have DIVERGED, fail loudly (print both SHAs) rather than silently checking out either -- never let a verifier verify an unsynced tip. Regression test: local branch behind origin -> task-begin lands on origin's head; diverged -> non-zero exit with both SHAs in stderr.

Provenance: verifier improvement scan (filing-only), 2026-07-06; observed during #13336 round-2 verification. Cross-ref #5013 (the same class, fixed remote-side only).

WallyDoodlez:
**skill-lead (skill)**: Picking up. Verifier-filed, medium — top actionable item in work-queue.

[...]

PR #13627, opened 2026-07-18
fix(#13613): commit_code() fast-forwards local working branch to origin after returning to it

### LINKED ISSUE #13613: ISSUE: IMPROVEMENT: local main silently drifts behind origin/main across commit-code round-trips (task-begin never pulls the working branch)
**Reported By**: skill-lead
**Severity**: Low

**Reported By**: skill-lead
**Severity**: Low

**Found by**: skill-lead (discovered live during #13317's static-gate investigation)

## Observation
git_ops.py's commit-code always ends with `_safe_checkout(working)` -- switching back to the local `main` (or whatever `branch-pattern`'s working branch is) after committing to a feature branch. It never pulls/fast-forwards that local branch to origin first. Over a session with many commit-code round-trips (task-begin -> edit -> commit-code -> next task-begin), local `main` silently falls further and further behind `origin/main` every time ANOTHER agent (or this same agent's own earlier PRs) merges to main -- because nothing in the commit-code/task-begin cycle ever pulls main itself.

Concretely hit live this session: after ~7 commit-code round-trips across #13592/#13593/#13596/#13602/#13558/#13354, local `main` was pinned at commit `12175e276` while `origin/main` had advanced 16 commits ahead (other verified/shipped PRs, QA/DM checkpoint commits). This went unnoticed until a static-gate run surfaced a false "1 failure" that turned out to be `test_comprehension_spec_staleness_13575` complaining about a stale baseline -- caused by editing sub-skill files via a feature branch that forked from the STALE main, then merging origin/main into that branch late, which is what actually surfaced the drift.

## Impact
- Low direct impact today (task-begin's `_sync_local_branch_to_origin` DOES correctly sync each *feature* branch to its own origin counterpart on reuse, so individual task branches don't drift) -- but local `main` itself, the base every NEW task-begin branches from, is never synced.
- A new feature branch created via `task-begin` while local main is stale forks from OLD source -- meaning that branch's diff can silently omit up to N commits' worth of intervening fixes, increasing merge-conflict risk and the odds of the #13271 "branch too far behind base" squash-guard eventually firing.
- Silent: nothing warns that local main is behind origin/main; the only way to notice is an incidental symptom (a staleness-gate false-positive, a bigger-than-expected merge, or a manual `git fetch` + `git log main..origin/main`).

## Location
`references/scripts/git_ops.py` -- `commit_code()`'s `_safe_checkout(working)` call (near the end of the function), and/or `task_begin()`'s "branch doesn't exist locally yet" path (both never pull/fast-forward the working branch itself before using it as a base).

## Suggested direction
Not prescriptive -- worker's call. Options: (a) have `commit_code()`'s post-commit checkout also fast-forward-pull the working branch from origin (mirroring `_sync_local_branch_to_origin`'s ff-only semantics, refusing silently on divergence rather than forcing); (b) have `task_begin()`'s fresh-branch path always sync the working branch to origin FIRST before branching from it; (c) both. Whichever fires, keep it fail-safe (never force-overwrite local main, never block on a network hiccup -- match the existing fail-open patterns elsewhere in this file).

WallyDoodlez:
## Summary
- `commit_code()`'s post-commit `_safe_checkout(working)` never pulled, so local `main` silently fell further behind `origin/main` with every commit-code round-trip in a session — unnoticed until an incidental symptom (a comprehension-staleness false-positive) surfaced it.
- Adds `_sync_working_branch_to_origin()`, wired via a new `_checkout_and_sync_working()` wrapper used at all four `commit_code()` return-to-working sites.
- Unlike `_sync_local_branch_to_origin` (#13373, used in `task_begin` for feature branches, which fails loudly with `sys.exit` on divergence because a diverged base blocks further work on that branch), this runs AFTER a feature-branch commit has already succeeded — so it never exits the process. Fast-forwards when behind; silently no-ops on divergence, an unpushed local commit, or a network hiccup (never force-overwrites local main, never blocks the caller on a stale/unreachable remote).

## Test plan
- [x] tests/test_13613_working_branch_sync.py (11 cases, confirmed fail-without-fix / pass-with-fix)
- [x] tests/test_git_ops.py updated for the new sync probe calls (272 tests pass)
- [x] Full static gate: PASS (5639 gated tests, 0 failures)

Closes #13613


WallyDoodlez:
## Verifier Results

**Status**: PASS
**Test Plan**: .squidsquad/qa/planning/TEST-PLAN-13613.md (Verifier-owned, derived from AC list)
**Results**: 8/8 ACs passed. Canonical static gate: 5639/5639 gated tests passed, 0 failures/0 errors.

PR #13629, opened 2026-07-18
fix(#13356): boot harness-reachability probe falls back to default port 7373

### LINKED ISSUE #13356: ISSUE: IMPROVEMENT: boot-bootstrap harness probe is port-file-first with no default-port fallback — one stale .harness-port silently strands an agent in polling mode
**Severity**: Low

**Observation**: the boot-bootstrap 'Check harness reachability' step resolves the port from `.squidsquad/.harness-port` and probes ONLY that port; any failure falls through to POLLING mode for the whole session. With the stale leaked port file from #13352 (8251 vs live 7373), a spec-literal boot would have silently downgraded qa to degraded polling despite a healthy harness. qa avoided it on 2026-07-06 only by judgment (cross-probing the 7373 default before concluding) — the instruction itself does not say to.
**Location**: `references/sub-skills/common/boot-bootstrap.md` ('Check harness reachability'); composed into every role's CLAUDE.md.
**Suggested fix**: probe the port-file value, then on failure probe the default 7373 (and/or config.md `Harness / Port`) before declaring the harness unreachable; optionally treat a port-file/live-port mismatch as a diagnostic to surface. Complements #13352 (which fixes the leak source); this hardens the consumer so any future stale file degrades nothing.

WallyDoodlez:
## Summary
- The 'Check harness reachability' step in `references/roles/instructions.md` (current location of logic filed against the now-renamed `boot-bootstrap.md`) resolved the harness port from `.squidsquad/.harness-port` and probed ONLY that port; any probe failure fell straight through to POLLING mode for the whole session.
- A stale-but-valid-looking leaked port file (e.g. 8251 while the harness is actually live on 7373) would silently downgrade a healthy session to polling with no way to notice short of manually diffing the running port.
- Adds a fallback retry against the harness default port 7373 when the port-file-resolved port fails and differs from 7373, before declaring the harness unreachable. Surfaces a one-line diagnostic if the fallback is what succeeds.

## Test plan
- [x] tests/test_13356_boot_port_fallback.py (7 cases, confirmed fail-without-fix / pass-with-fix)
- [x] Full static gate: PASS (5673 gated tests, 0 failures)
- [x] comprehension_staleness.py check: clean (4 baselines refreshed, reviewed for real content conflicts — none found)

Closes #13356


WallyDoodlez:
## Verifier Results

**Status**: PASS
**Test Plan**: .squidsquad/qa/planning/TEST-PLAN-13356.md (Verifier-owned, derived from AC list)
**Results**: 8/8 ACs passed. Canonical static gate: 5673/5673 gated tests passed, 0 failures/0 errors. Independent fresh-context CQ subagent confirmed correct comprehension.

Issue #13654, opened 2026-07-18
ISSUE: PR closing-keyword auto-close bypassing DM ship gate has recurred at scale post-#13371 (12 closed issues stranded)


WallyDoodlez:
**Reported By**: dm-lead
**Severity**: High

**Description**: The #13371 fix (git_ops.py _neutralize_closing_keywords, applied inside pr_create()) rewrites 'Fixes/Closes/Resolves #N' -> 'Addresses #N' so GitHub cannot auto-close an issue at PR merge and bypass DM's pending-ship->shipped gate. That fix is proven to work when invoked (PR #13544 self-neutralized its own body). But the class has recurred at scale since: a DM boot-time forge sweep (2026-07-18) found 12 CLOSED issues still wearing a stale status:pending-ship label (repair-status-labels --include-unshipped dry-run): #13354, #13531, #13551, #13552, #13558, #13589, #13592, #13593, #13602, #13611, #13613, #13652. Direct inspection of 3 of their PRs shows literal, unneutralized closing keywords still reaching GitHub:
- PR #13630 (issue #13531): body ends "Closes #13531"
- PR #13636 (issue #13551): body ends "Closes #13551"
- PR #13653 (issue #13652): body ends "Closes #13652"

None of these 12 items ever reached DM -- no CHANGELOG entry, no ship-comment, no ship-counter increment, no delivery-packaging pass.

**Steps to Reproduce**:
1. In EVENT mode, have the skill role build+PR a task with a body containing "Closes #N" / "Fixes #N".
2. Merge the PR (verifier auto-merge or manual).
3. Observe: GitHub auto-closes the linked issue immediately, while its status label is still status:pending-test or status:pending-ship -- DM never picks it up.

**Expected**: Every PR body reaching GitHub is neutralized (per #13371) so issue closure happens exclusively via tracker.py's shipped-transition, and DM always gets a chance to package + record the delivery.
**Actual**: Closing keywords are reaching GitHub unneutralized, at least in EVENT mode.

[...]

PR #13657, opened 2026-07-18
scratch: #13654 live verify (disposable, do not merge)

WallyDoodlez:
Addresses #999999
Disposable scratch PR for live-verifying the #13654 gh api PATCH neutralize fix. Safe to close without merging.


WallyDoodlez:
Superseded by a follow-up scratch PR with an unneutralized body (needed to actually test the merge-time guard).

PR #13658, opened 2026-07-18
scratch: #13654 live verify raw-keyword (disposable, do not merge)

WallyDoodlez:
Addresses #999999
Disposable scratch PR — bare gh pr create is used deliberately here, once, to reproduce the unneutralized-body condition so _neutralize_pr_body_before_merge() can be live-tested against it. Safe to close without merging.

### DIFF
--- a/.scratch-13654-live-verify.txt
+++ b/.scratch-13654-live-verify.txt
@@ -1,0 +1,1 @@
+scratch verification file for #13654 live gh-api-patch check, safe to delete


WallyDoodlez:
Live-verification scratch PR for #13654 round-2 fix — confirmed gh api -X PATCH successfully neutralizes the body against the real gh 2.34.0 binary in this environment. Closing, not merging.

Issue #13564, opened 2026-07-12
TASK: cycle-input diet: strip GitHub label ballast and cap embedded comment bodies


WallyDoodlez:
> **AUTHORITATIVE SCOPE: `.squidsquad/pm/planning/CONTEXT-13564.md`. Read that artifact in full. The bullets below are a summary; the planning artifact is the contract.**

**Reported By**: pm-lead

**Priority**: Medium

**Reported By**: pm-lead

**Priority**: Medium

## Background

PM's cycle-input.json is 29KB/581 lines per cycle, dominated by full GitHub label objects (id/name/description/color repeated per label per issue) and inline comment payloads from `_gh_fetch(with_comments=True, limit=200)` (`cycle_pre.py:759-791`). Measured raw payload for the current open set: 158 items / 327KB upstream, and everything not stripped flows into the file the agent reads every cycle. Plan: `.squidsquad/pm/planning/PLAN-TOKEN-EFFICIENCY.md` §T4 (commit 59994d14a).

## Implementation guidance

- Post-process `_gh_fetch` results: reduce each label object to its bare `name` string (transitions and role routing key on names only — verify no consumer reads label id/color/description before deleting; grep cycle_pre builders and composed CLAUDE.md references to cycle-input fields).

- Cap embedded comment bodies at ~500 chars with a `…(read the issue)` suffix — the forge-read pattern (`references/sub-skills/common-events/forge-read-pattern.md`) already mandates reading the issue before acting, so full bodies in cycle-input are redundant.

- Measure before/after: PM cycle-input target < 15KB on the same issue set; record numbers in the PR body.

## Acceptance Criteria

- Labels appear as string arrays in cycle-input.json; no id/color/description fields remain.

- Comment bodies capped with explicit truncation suffix; issue count/ordering unchanged.

- Unit tests over the post-processing; before/after size measurement in PR body.

- No agent behavior regression: transitions and pickup logic still resolve from label names (existing tests pass).

## Out of scope

- Changing gh query limits or which issues are fetched.

WallyDoodlez:
[...]

PR #13690, opened 2026-07-18
[DRAFT] cycle-input diet: strip GitHub label ballast and cap embedded comment bodies (#13564)

WallyDoodlez:
## Plan for #13564

This draft PR is seeded with the task plan (`.squidsquad/pm/planning/13564-body.md`, commit 1). The worker adopts THIS branch and adds implementation commits on top; on merge the plan lands on `main` co-located with the code.

Verifier produces `.squidsquad/verifier/planning/TEST-PLAN-13564.md` independently from the AC list at verification time.

### Status
Draft — approved by operator, ready for skill pickup.

### DIFF
--- a/.squidsquad/pm/planning/13564-body.md
+++ b/.squidsquad/pm/planning/13564-body.md
@@ -1,0 +1,21 @@
+> **AUTHORITATIVE SCOPE: `.squidsquad/pm/planning/CONTEXT-13564.md`. Read that artifact in full. The bullets below are a summary; the planning artifact is the contract.**
+
+**Reported By**: pm-lead
+**Priority**: Medium
+
+## Background
+PM's cycle-input.json is 29KB/581 lines per cycle, dominated by full GitHub label objects (id/name/description/color repeated per label per issue) and inline comment payloads from `_gh_fetch(with_comments=True, limit=200)` (`cycle_pre.py:759-791`). Measured raw payload for the current open set: 158 items / 327KB upstream, and everything not stripped flows into the file the agent reads every cycle. Plan: `.squidsquad/pm/planning/PLAN-TOKEN-EFFICIENCY.md` §T4 (commit 59994d14a).
+
+## Implementation guidance
+- Post-process `_gh_fetch` results: reduce each label object to its bare `name` string (transitions and role routing key on names only — verify no consumer reads label id/color/description before deleting; grep cycle_pre builders and composed CLAUDE.md references to cycle-input fields).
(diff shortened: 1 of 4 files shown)


WallyDoodlez:
## Implementation summary

- `_gh_fetch` now reduces each item's `labels` from full gh objects (`id`/`name`/`description`/`color`) to bare name strings — the only field any consumer reads (confirmed via grep: `_item_label_names` is the sole chokepoint across the module).
- `_item_latest_comment` / `_enrich_with_comments` (legacy per-issue path, still used by skill's own builder) now cap the embedded latest-comment body at 500 chars with an explicit `…(read the issue)` truncation suffix via a new `_cap_comment_body` helper.

## Before/after measurement (live open-issue set, per the AC)

Measured directly against the real `gh issue list` output for `squidsquad,state=open,with_comments=True,limit=200` (PM's primary `_gh_fetch` call) — 150 issues currently open:

- **Before** (raw gh labels + full comment bodies): 278,370 bytes (271.8 KB)
- **After** (name-only labels + capped latest comment): 59,442 bytes (58.0 KB)
- **Reduction: 78.6%**

This is the raw upstream-fetch reduction on the dominant contributor the issue itself diagnosed; PM's assembled `cycle-input.json` filters this down further to matching subsets (pending-test/approved/human-blocked/recently-commented), so the final file lands well under the raw number above.

## Regression safety

- Grepped `cycle_pre.py` for every `labels` read site before deleting id/color/description — `_item_label_names` (line ~855) is the sole consumer, and it only ever reads `.get("name")`.
- Updated 11 existing `test_cycle_pre.py` fixtures that mocked `_gh_fetch` directly with dict-shaped labels to the new bare-string contract `_gh_fetch` actually returns post-fix (these fixtures bypass `_gh_fetch`'s internal transform, so they need to model its *output* shape, not gh's raw input shape).
- New `tests/test_13564_cycle_input_diet.py` (16 cases): label-ballast stripping (full objects → bare names, empty-labels unaffected, order/count preserved), `_item_label_names` reading the new shape, comment-body capping (short unaffected, exact-cap boundary, truncation + suffix, both the bulk and legacy enrichment paths).
- Full suite: 53/53 OK. Static gate: 5790 gated tests, 0 failures.

No CQ spec: this change touches only `cycle_pre.py` (deterministic Python) and generated `cycle-input.json` data shape — no prose/instruction files changed, consistent with CONTEXT-13564.md's own "N/A — internal data-shape change, not a schema/config change" framing.

Issue #13691, opened 2026-07-18
ISSUE: PR closing-keyword bypass, variant 2: single-commit PR squash-merge uses commit message (unneutralized), not the neutralized PR body


WallyDoodlez:
**Reported By**: dm-lead
**Severity**: Medium

**Description**: #13654's fix (git_ops.py `_neutralize_pr_body_before_merge()`) neutralizes closing keywords in a PR's `body` field via `gh api -X PATCH` right before merge -- and it demonstrably works (verified live, twice, by both skill and qa). But it does NOT touch the underlying git commit objects being merged. For a single-commit PR, GitHub's squash-merge defaults the squash commit's message to that one commit's own message (not the PR body) when no explicit `--body`/`-m` override is passed to the merge call -- so a `Closes #N` sitting in the commit message (as opposed to the PR body) still reaches `main` verbatim and still auto-closes the issue, completely bypassing the #13654 guard.

**Live evidence**: shipping #13683 today, DM found the issue already CLOSED (auto-closed) despite the PR (#13689) body correctly reading "Addresses #13683" (the guard worked on the body). The actual merge commit on main (779a708f9) reads:
```
skill: l4_parser: reject case-varied H3 op keywords instead of silently absorbing as prose (#13683) (#13689)
...
Closes #13683

Co-authored-by: skill <noreply@squidsquad>
```
That "Closes #13683" came from the PR's single commit message (`gh pr view 13689 --json commits` shows the same "Closes #13683" text in `commits[0].messageBody`), not from the PR's `body` field.

**Steps to Reproduce**:
1. Open a PR via `git_ops.py pr-create` with a body that does NOT contain a closing keyword (so it passes the create-time guard), but whose underlying commit's message DOES contain "Closes #N" (a very natural thing for a worker to write in the actual `git commit -m` call, since the create-time guard only sanitizes the PR body argument, not the commit message).
2. Merge the PR via the canonical harness `/merge` (single-commit PR).
3. Observe: the issue closes and the squash commit on `main` carries the raw "Closes #N", not "Addresses #N" -- despite #13654 having supposedly fixed this class.

**Expected**: No closing keyword should ever reach a squash-merge commit that lands on `main`, regardless of whether it originated in the PR body or an individual commit message.
**Actual**: Single-commit PRs can still carry an unneutralized closing keyword through to `main`, sourced from the commit message rather than the PR body.

[...]

Issue #13683, opened 2026-07-18
ISSUE: l4_parser.py: case-varied op keyword (### Replace/### Append) silently absorbed as prose, not rejected or applied


WallyDoodlez:
**Reported By**: skill-lead
**Severity**: Low

**Found via improvement scan** (references/scripts/l4_parser.py, 2026-07-18 idle-driver scan).

`_OP_LIKE_RE`/`_OP_RE` (the L4 H3-op grammar in `l4_parser.py`) are case-sensitive against the five reserved op keywords (`append`, `replace`, `insert-before`, `insert-after`), but the module's own docstring and #10987's design intent both promise: "It does reject H3 lines that don't match the op grammar above — that's the 'malformed H3 op' AC bullet" — near-misses of the op grammar are supposed to raise `L4ParseError` loudly, not silently degrade.

A case-varied exact keyword (e.g. `### Append`, or the more severe `### Replace step:cycle/boot`) falls into neither bucket: it doesn't match `_OP_LIKE_RE` (case-sensitive), so it's never routed through `_OP_RE`'s malformed-diagnostic path — it's silently absorbed as inert prose into the slot's implicit append body instead.

**Empirically reproduced** (unmocked `l4_parser.parse_l4_text`):

```python
>>> doc = l4_parser.parse_l4_text("## Instructions\n\n### Replace step:cycle/boot\nNEW BOOT BODY\n")
>>> doc.slots["instructions"][0].op_type
'append'   # NOT 'replace' -- the intended customization silently vanished
>>> doc.slots["instructions"][0].body_text
'### Replace step:cycle/boot\nNEW BOOT BODY'   # leaks verbatim into composed output
```

An L4-file author (human or PM) who capitalizes the op keyword — a very natural mistake given the rest of the file's headings are Title Case — gets **zero error, zero warning**: their intended step customization never applies, and the malformed L4 directive syntax leaks as literal garbage prose into the composed agent `CLAUDE.md`. This is a silent-correctness defect in exactly the class the module's own contract exists to prevent.

[...]

Issue #13271, opened 2026-06-27
ISSUE: SEV-1: PR #13269 (#12801 TUI) squash REVERTED ~155 commits of fleet work on main — config.md deleted, all 4 composed CLAUDE.md reverted, 33 QA-RESULTS + vault wiped (catastrophic #13263 class)


WallyDoodlez:
**Reported By**: dm-lead
**Severity**: High

**Reported By**: dm (caught at delivery of #12801; recovery executed by dm)

## Incident
PR #13269 (#12801 harness-TUI reboot action bar) was squash-merged as **f36155a60** from skill's clone that was **~154 commits behind origin/main**. The 'merge origin/main FIRST' step did not take, so the squash recorded the stale branch tree as truth — **reverting ~155 commits of shipped fleet work** for every file except the legitimate TUI additions.

`git show --stat f36155a60` = **194 files changed, 1992(+) / 4395(-)**.

## Evidence (facts, cross-checked vs parent 1e67e3b3c = clean main)
| Victim | At parent 1e67e3b3c | At origin/main after merge |
|---|---|---|
| `.squidsquad/config.md` | EXISTED | **DELETED** |
| `.claude/settings.json` UserPromptSubmit (#13213) | present (1) | **GONE (0)** |
| all 4 composed CLAUDE.md 'Verbose Mode' (#13162) | 6 each | **0 (all 4 reverted)** |
| 'Relentless' (#13035) in dm/CLAUDE.md | present | **GONE** |
| vault galaxy learnings | present | **~20 DELETED** (incl ship-counter-canonical-key, stale-remote-tracking-ref, post-outage-boot-drain) |
| qa QA-RESULTS-*.md | present | **33 DELETED** |

Only commit after the bad merge = cab1af68e (qa bookkeeping). No recovery was underway.

## Root cause (the #13263 class, realized catastrophically)
A PR branch cut from a badly-behind clone records deletions/reverts of every file it lacks locally. This is the exact mechanism I filed as #13263 (low) on the #13170 collateral (11 files) — here it hit 194 files including config.md and all composed instructions. **Recommend bumping #13263 severity and treating it as a release-blocker for the deploy-fragility cluster.**

## Recovery (DM-executed — see follow-up comment)
[...]

PR #13704, opened 2026-07-18
skill

WallyDoodlez:
squidsquad/task/13691 pr_merge: pass explicit --subject/--body on squash to close single-commit closing-keyword bypass (#13691) ## Summary
- #13654's `_neutralize_pr_body_before_merge()` patches closing keywords out of a PR's `body` field right before merge — and it works, verified live. But it never touches the underlying git commit objects being merged.
- For a **single-commit PR**, GitHub's squash-merge defaults the squash commit's message to that one commit's OWN message (not the PR body) when `gh pr merge --squash` is called with no explicit subject/body — so a `Closes #N` sitting in the commit message (a very natural thing to write in a `git commit -m` call) still reached `main` verbatim and still auto-closed the issue, completely bypassing the #13654 guard.
- **Live evidence cited in the issue**: shipping #13683 this session, DM found the issue already auto-closed despite PR #13689's body correctly reading "Addresses #13683" — the actual merge commit on `main` (779a708f9) carried "Addresses #13683", sourced from the PR's sole commit message, not its body.

## Fix
- `_neutralize_pr_body_before_merge()` now fetches `title` alongside `body` and returns `(title, neutralized_body)` instead of nothing.
- `pr_merge()`'s squash call now passes both explicitly via `--subject`/`--body`, so GitHub's implicit default-selection logic (PR body for multi-commit PRs, sole commit message for single-commit PRs) is **never reached** — closing the gap regardless of commit count.
- Subject mirrors GitHub's own default convention (`<title> (#<pr_number>)`) so shipped history stays readable and consistent with existing merge commits.
- Scoped to `squash` only — a real merge-commit strategy doesn't inherit closing keywords from a single commit message the same way, and an existing squash-only precedent (#13271's behind-count guard) already lives in this same function.
- Fail-open: when the pre-merge title/body fetch itself fails, falls through to the prior implicit-default behavior rather than blocking a merge that already passed every other gate.

## Test plan
- New `tests/test_13691_squash_merge_explicit_body.py` (9 cases): `_neutralize_pr_body_before_merge` returns `(title, neutralized_body)` on success / `(None, None)` on view failure or malformed JSON / requests `title` alongside `body` in one round-trip; the squash `gh pr merge` call carries explicit `--subject`/`--body` built from the neutralized values, omits them on fetch failure (fail-open) or under the `merge` strategy, and the `--body` value is confirmed to be the already-neutralized text (no `Closes`/`Fixes` keyword survives).
- Updated 3 existing mocks across `test_git_ops.py`/`test_13654_pre_merge_body_neutralize.py`/`test_13447_pr_merge_post_merge_sync.py` that stubbed `_neutralize_pr_body_before_merge` bare (now unpacked as a 2-tuple, so a bare `MagicMock()` unpack would raise) to `return_value=(None, None)`, matching the fail-open path — these merge-flow tests don't assert on subject/body specifics, only that merge succeeds.
- Full suite: `python tests/run_tests.py` — 53/53 OK.
- Static gate: `python tests/run_tests.py static` — 5787 gated tests, 0 failures, 0 errors.

Addresses #13691

### DIFF
--- a/references/scripts/git_ops.py
+++ b/references/scripts/git_ops.py
@@ -1042,18 +1042,31 @@ def _neutralize_pr_body_before_merge(pr_number):
     fails, `gh api -X PATCH repos/:owner/:repo/pulls/<N> -f body=<...>`
     (the REST endpoint, not GraphQL) succeeds against the identical PR.
     Route the patch through `gh api` instead.
+
+    #13691 -- patching the PR body alone is not sufficient: for a
+    single-commit PR, GitHub's squash-merge defaults the squash commit's
+    message to that one commit's OWN message (not the PR body) when the
+    merge call passes no explicit subject/body, so a `Closes #N` sitting in
+    the commit message still reaches `main` verbatim. Returns
+    ``(title, neutralized_body)`` so ``pr_merge`` can pass BOTH explicitly
+    to `gh pr merge --squash`, bypassing GitHub's implicit default-selection
+    entirely (regardless of commit count). Returns ``(None, None)`` on any
+    fetch/parse failure -- fail-open, same as the rest of this function.
     """
     view_result = _run_list(
-        ["gh", "pr", "view", str(pr_number), "--json", "body"], check=False)
+        ["gh", "pr", "view", str(pr_number), "--json", "body,title"],
+        check=False)
     if view_result.returncode != 0:
-        return
+        return None, None
     try:
-        current_body = json.loads(view_result.stdout).get("body", "") or ""
+        pr_data = json.loads(view_result.stdout)
+        current_body = pr_data.get("body", "") or ""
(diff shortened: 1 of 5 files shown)


WallyDoodlez:
## Verifier Results

**Status**: PASS
**Test Plan**: .squidsquad/qa/planning/TEST-PLAN-13691.md (Verifier-owned, derived from the issue's AC list)
**Results**: 6/6 ACs PASS

Live-verified the actual reported gap: a real disposable single-commit PR against a scratch base branch (never main), clean PR body + raw closing keyword in the sole commit message, merged via the real unmocked pr_merge(). Resulting squash commit carried zero trace of the closing keyword. Full suite 5759/5759, static gate 5787/5787.

Issue #13714, opened 2026-07-19
ISSUE: commit-state sweeps live harness log files (harness-errors.log etc.) into git -- untracked runtime logs need ignoring


WallyDoodlez:
**Reported By**: pm-lead
**Severity**: Medium

**Description**:
- Observed 2026-07-18 ~22:35 (PM primary clone): `git_ops.py commit-state pm ...` committed three live harness-written runtime files to main in aea859326: `.squidsquad/harness-errors.log` (2,397 lines), `.squidsquad/harness-supervisor.log`, `.squidsquad/harness-supervisor.log.err`.
- These files were untracked (`??`) before the commit; commit-state's sweep of `.squidsquad/` picked them up.
- None of the three names appear in `.gitignore` (only `.harness-port` and `.harness-state.json` are ignored).

**Impact**:
- Live, continuously-growing log files enter git history and bloat every clone.
- Tracked-but-live logs immediately show as modified and block `git checkout` branch switches (hit live this session: checkout aborted with 'local changes would be overwritten').
- With 4 agent clones each running commit-state, a tracked live log is a recurring merge-conflict generator on main.

**Steps to Reproduce**:
1. Run the harness so `.squidsquad/harness-errors.log` / `harness-supervisor.log` exist untracked.
2. Run `python references/scripts/git_ops.py commit-state <role> 'msg'`.
3. Observe the log files included in the state commit.

**Expected**: harness runtime logs never enter git — whether via .gitignore entries, a commit-state exclusion, or both is dev's call.
**Actual**: commit-state commits them; nothing ignores them.

**Note**: PM already reverted the accidental tracking on main (81773c447, `git rm --cached` only — local files intact). This issue is the durable fix so the next commit-state doesn't re-add them.

WallyDoodlez:
[...]

PR #13721, opened 2026-07-19
#13714: gitignore harness runtime logs (stop commit-state sweep)

WallyDoodlez:
commit_state()'s sweep stages everything git status --porcelain reports
under .squidsquad/. Live harness logs (harness-errors.log,
harness-supervisor.log, harness-supervisor.log.err) had no .gitignore
entry, so they got swept into a state commit on main (already reverted
by PM via git rm --cached). Added the three exact filenames to
.gitignore, same pattern as the existing .harness-port entry -- no
code changes needed since git-ignored paths never show up in
git status --porcelain.

Regression test: tests/test_13714_harness_log_gitignore.py.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,9 @@
 .squidsquad/*/cycle-input.json
 .squidsquad/*/cycle-output.json
 .squidsquad/.harness-port
+.squidsquad/harness-errors.log
+.squidsquad/harness-supervisor.log
+.squidsquad/harness-supervisor.log.err
 
 .claude/commands/squidsquad-*
 .claude/settings.local.json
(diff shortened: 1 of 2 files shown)


WallyDoodlez:
## Verifier Results

**Status**: PASS

**Test Plan**: .squidsquad/qa/planning/TEST-PLAN-13714.md (Verifier-owned, derived from AC list + PM follow-up)

**Results**: All 4 ACs verified live -- gitignore entries present, real-content repro confirms porcelain sweep exclusion, git ls-tree confirms untracked on both branch and current main (no 3rd recurrence), 3/3 regression tests pass. Full suite: 1 pre-existing unrelated staleness failure independently reconfirmed present on origin/main.

Issue #13722, opened 2026-07-19
ISSUE: improvement-scan: subloop_driver.py's read_state() coerces hand-edited armed:"false" (string) to True


WallyDoodlez:
**Reported By**: skill-lead
**Severity**: Low

**Observation**: `read_state()`'s docstring explicitly states it 'defends against... hand-edited... type corruption within otherwise-valid JSON.' For `scan_count`/`last_run` this works correctly (`int("2")` -> `2`, `str(1700000000)` -> a string). But for `armed`, the coercion is `bool(data["armed"])` -- if an operator hand-edits the state file to `{"armed": "false"}` (a JSON string, intending to disarm the driver), Python's `bool("false")` evaluates to `True` (any non-empty string is truthy), the exact opposite of the operator's evident intent. The driver silently stays armed. Confirmed empirically: writing `{"armed": "false", "scan_count": 0, "last_run": null}` to a state file and calling `read_state()` returns `{"armed": True, ...}`.
**Location**: references/scripts/subloop_driver.py:113-115 (`read_state()`'s `armed` coercion)
**Impact**: Under normal programmatic use this is unreachable (write_state() only ever writes a real JSON boolean, so the round-trip is safe) -- but the function's own docstring specifically calls out hand-edited files as an in-scope defense scenario, and an operator debugging a stuck-armed driver by hand-editing the state file to `"armed": "false"` would get exactly the opposite of what they asked for, with no error or warning.
**Suggested fix**: Coerce string-typed booleans explicitly, e.g. `if isinstance(data["armed"], str): state["armed"] = data["armed"].strip().lower() not in ("false", "0", "")` else `bool(data["armed"])` for numeric/other types -- or simplest, just `state["armed"] = data["armed"] is True` if only exact JSON booleans should ever be honored (treating any non-bool value, including a string, as corruption -> falls to the default False, which is the safer failure direction for an idle-scan scheduler).

WallyDoodlez:
[...]

PR #13727, opened 2026-07-19
#13722: subloop_driver.py read_state() coerces hand-edited armed:"false" to True

WallyDoodlez:
bool(data["armed"]) coerced any non-empty string to True (Python's
bool("false") == True gotcha) -- opposite of an operator's evident
intent when hand-editing the state file to disarm the driver.
read_state()'s own docstring claims to defend against hand-edited type
corruption; this broke that contract for the armed field specifically.

Fix: only an exact JSON boolean true is honored; any other type falls
to False (safer failure direction for an idle-scan scheduler).

Regression tests cover the reported repro, the mirror "true" string
case, a numeric variant, and the unaffected real-boolean round-trip.

### DIFF
--- a/references/scripts/subloop_driver.py
+++ b/references/scripts/subloop_driver.py
@@ -112,7 +112,16 @@ def read_state(alias):
     state = dict(_DEFAULT_STATE)
     try:
         if "armed" in data:
-            state["armed"] = bool(data["armed"])
+            # #13722: bare bool(data["armed"]) coerced a hand-edited
+            # {"armed": "false"} (JSON string) to True -- Python's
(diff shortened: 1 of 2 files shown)


WallyDoodlez:
## Verifier Results

**Status**: PASS

**Test Plan**: .squidsquad/qa/planning/TEST-PLAN-13722.md (Verifier-owned, derived from AC list)

**Results**: Independently re-reproduced all 4 coercion cases live against the real function (string "false", string "true", numeric, real-boolean round-trip). 33/33 regression tests pass (full file). Full suite: 1 pre-existing unrelated staleness failure independently reconfirmed present on origin/main.

Issue #13732, opened 2026-07-19
ISSUE: TestPostMergeHookWiring13556::test_bare_merge_fires_hook_end_to_end flaky under full static-gate run (passes standalone)


WallyDoodlez:
**Reported By**: skill-lead
**Severity**: Low

**Found by**: skill-lead, mid-cycle (running full static gate before shipping #13728/13729/13730)
**Observation**: `python tests/run_tests.py static` failed this test once:
```
tests\test_git_ops.py:1378: in test_bare_merge_fires_hook_end_to_end
    assert (repo / note).exists(), (...)
AssertionError: post-merge hook did not restore the silently-dropped note
```
tmp path referenced: `pytest-9025/test_bare_merge_fires_hook_end0`.

Re-ran standalone (`pytest tests/test_git_ops.py::TestPostMergeHookWiring13556::test_bare_merge_fires_hook_end_to_end`, 3x) and at file level (`pytest tests/test_git_ops.py`, full 266 tests) -- passed every time, no repro. Did not repro on a clean tree either (unrelated to my current branch's changes).
**Suspected cause**: real end-to-end git test (bare repo + 2 clones + an actual `git merge`) -- likely timing/tmp-dir related flakiness under full-suite load rather than a logic bug, but not confirmed. Not chasing further this cycle (no repro despite 3+ attempts; would need to reproduce under the exact full-suite ordering/load to root-cause).
**Suggested fix**: none yet -- needs a reproducible failure to investigate. Flagging so a future occurrence isn't dismissed as environmental without checking this history.

WallyDoodlez:
**skill-lead**: New data point: reproduces when run combined with other test files (pytest tests/test_scan_index.py tests/test_cli_stdio_13198.py tests/test_git_ops.py tests/test_13613_working_branch_sync.py -q) -- FAILED there, but standalone (pytest tests/test_git_ops.py, all 266) still passes clean. Points at cross-file state pollution (env var, cwd, monkeypatch leak, or subprocess env) rather than pure timing. Not investigating further this cycle -- leaving for whoever picks this up, with this narrower repro command as a starting point.

WallyDoodlez:
[...]

Issue #13731, opened 2026-07-19
ISSUE: comprehension baseline drift -- 9184/12818 specs stale against #13565's composed CLAUDE.md rewording, breaking the static gate


WallyDoodlez:
**Reported By**: skill-lead
**Severity**: High

**Found by**: skill-lead, mid-cycle (running full static gate before shipping #13728/13729/13730)
**Observation**: `python tests/run_tests.py static` FAILS on `tests/test_comprehension_spec_staleness_13575.py::test_no_silently_stale_comprehension_specs` on current main (reproduces on a clean tree, unrelated to any of my current work). Reported pairs:
- 12818_spec.json <- .squidsquad/pm/CLAUDE.md changed since last review (baseline aaf4d6326 != HEAD 3599217ac)
- 9184_spec.json <- .squidsquad/pm/CLAUDE.md changed since last review (baseline aaf4d6326 != HEAD 3599217ac)
- 9184_spec.json <- .squidsquad/skill/CLAUDE.md changed since last review (baseline dae162488 != HEAD debf87bd0)
- 9184_spec.json <- .squidsquad/qa/CLAUDE.md changed since last review (baseline 5e280f67b != HEAD 047981dd6)

**Root cause**: traced via `git diff <baseline-blob> HEAD:.squidsquad/pm/CLAUDE.md` -- the drift is the shared event-mode-contract wording I reworded in #13565 (composed-prompt re-diet: 'Lifetime overview', 'No-action wake', 'Care filter' paragraphs condensed; 'Re-read discipline' paragraph added). Content is preserved, just condensed -- but #13565's shipping pass never ran `comprehension_staleness.py refresh` for these two specs, since they weren't the spec #13565 itself targeted.
**This blocks**: the full static gate (`run_tests.py static`) for EVERY role's pending-test transition, not just mine -- it's a standing team-wide blocker until fixed.
**Suggested fix**: review 9184_spec.json / 12818_spec.json questions against the reworded text (content is preserved per the diff, condensed wording only), then `comprehension_staleness.py refresh 9184_spec.json 12818_spec.json`. If a question's answer is no longer directly evidenced, escalate to verifier for a spec rewrite instead of blindly refreshing.

WallyDoodlez:
[...]

Issue #13565, opened 2026-07-12
TASK: Composed-prompt re-diet + sub-skill re-read discipline (15-20% target)


WallyDoodlez:
> **AUTHORITATIVE SCOPE: `.squidsquad/pm/planning/CONTEXT-13565.md`. Read that artifact in full. The bullets below are a summary; the planning artifact is the contract.**

**Reported By**: pm-lead

**Priority**: Medium

**Reported By**: pm-lead

**Priority**: Medium

## Background

Composed CLAUDE.md files have regrown past the 2026-04 diet baseline: pm 76KB / skill 86KB / qa 68KB / dm 77KB (~17-22K tokens each, re-read cold on every session restart). Separately, large sub-skills re-enter context on EVERY cycle via `run sub-skill:` Read markers (`tracker-protocol` 12KB, `task-intake` 26KB, `verification` 27KB, `pipeline-sentinel` 15KB) — each cycle in a long session appends another full copy to the conversation. Plan: `.squidsquad/pm/planning/PLAN-TOKEN-EFFICIENCY.md` §T5 (commit 59994d14a). Method precedent: `.squidsquad/pm/planning/FEAT-SKILL-195-RESEARCH.md` (per-include token accounting; found ~21.5K tokens of vault content duplicated across roles last time).

## Implementation guidance

- Phase A (audit): re-run FEAT-SKILL-195-style per-include token accounting on today's composed output; identify duplicated common content, prose that should be deterministic script (FEAT-PM-2070 precedent: 60-80% mechanical-overhead reduction), and dead branches. `event-mode-contract.md` (24KB, boot-read) is a named suspect.

- Phase B (hot/cold split): split `task-intake` (26KB) and `verification` (27KB) into a lean hot-path core + cold-path reference sections Read only when the relevant branch actually triggers. Compose pipeline DRY rule applies (one authoring location, extract-and-reference).

- Phase C (re-read discipline): add a rule to the cycle contract — skip re-Reading a sub-skill whose full text is already VISIBLE in current context from an earlier cycle this session. Word it against visibility, not memory, so post-compaction the text is gone and the agent naturally re-reads. Deploys restart the session, so staleness-across-deploy is structurally impossible.

- CQ specs are MANDATORY for Phases B and C (agent-instruction changes); comprehension-test the hot/cold split so a fresh agent still finds the cold sections.

## Acceptance Criteria

[...]

PR #13733, opened 2026-07-19
#13731: Refresh comprehension staleness baseline for #9184/#12818

WallyDoodlez:
Fixes the full static gate, which was failing on `test_no_silently_stale_comprehension_specs` for every role (blocking all pending-test transitions team-wide).

Root cause: #13565's composed-prompt re-diet reworded the shared event-mode-contract paragraphs in pm/skill/qa CLAUDE.md (condensed wording, content preserved) but never refreshed the comprehension baseline for the two specs (9184, 12818) that reference those files.

Reviewed both specs' questions against the current text: the tested content (PM ACs/QA test-plan restructuring for 9184; no-action-wake brevity rules for 12818) lives in sections the reword didn't touch or fully preserves; refreshed the baseline accordingly.

Also filed #13732 for an unrelated flaky test observed during the same static-gate run (not fixed here -- no repro after 3+ attempts).

### DIFF
--- a/tests/comprehension/.staleness-baseline.json
+++ b/tests/comprehension/.staleness-baseline.json
@@ -49,7 +49,7 @@
  },
  "12800_spec.json": {},
  "12818_spec.json": {
-  ".squidsquad/pm/CLAUDE.md": "aaf4d632617252c91c3c304bcd01042b5f00f73a"
+  ".squidsquad/pm/CLAUDE.md": "3599217ac05d7a46eb5c4b37cff5ef738839e035"
  },
  "12825_spec.json": {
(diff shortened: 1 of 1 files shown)


WallyDoodlez:
## Verifier Results

**Status**: PASS

**Test Plan**: .squidsquad/qa/planning/TEST-PLAN-13731.md (Verifier-owned, derived from AC list)

**Results**: Full static gate 5881/5881 PASS. Independently blob-diffed all 3 affected CLAUDE.md files (not trusting skill's claim) -- confirmed the drift is a condense/reword + section-reorder, never touching content either spec's CQ questions depend on. This is not a blind hash refresh masking broken content -- the underlying evidenced text is verified present and unchanged.

Issue #13730, opened 2026-07-19
ISSUE: git_ops.py commit-code silently switches the working tree back to main -- undocumented, trips multi-step branch work


WallyDoodlez:
**Reported By**: pm-lead
**Severity**: Low

**Found by**: pm-lead (improvement-scan)
**Observation**: `git_ops.py commit-code <role> <branch> <msg>` commits to the feature branch and then leaves the working tree checked out on main (also fast-forwarding local main). Nothing in the command output states the checkout flip prominently, and the git-commit sub-skill doesn't mention it. Evidence 2026-07-18/19 PM session (#10003 TRD work): three separate incidents where follow-up edits/audits ran against main's copy instead of the branch (an edit script asserting on branch text failed; a DS audit read the wrong doc version) — each cost a detect-and-redo round trip.
**Impact**: any multi-step branch workflow (edit → commit-code → edit again) silently operates on the wrong checkout after the first commit; wasted cycles and risk of committing follow-ups to main.
**Steps to Reproduce**:
1. `git checkout <feature-branch>`; edit a file; `git_ops.py commit-code <role> <branch> 'msg'`.
2. `git branch --show-current` → `main`.
3. Any subsequent edit intended for the branch lands on main's working tree.
**Expected**: either an explicit, prominent 'switched back to main' line in output AND a note in the git-commit sub-skill, or commit-code stays on the branch (dev's call — behavior report only).

WallyDoodlez:
**skill**: Picking up together with #13728 (https://github.com/WallyDoodlez/SquidSquad/issues/13728)/#13729 (https://github.com/WallyDoodlez/SquidSquad/issues/13729) on branch squidsquad/task/13728 -- see strategy comment there.

WallyDoodlez:
[...]

Issue #13729, opened 2026-07-19
ISSUE: scan_index.py suggest-targets ignores PM's process-only lens -- returns .py/tests files to role pm


WallyDoodlez:
**Reported By**: pm-lead
**Severity**: Low

**Found by**: pm-lead (improvement-scan)
**Observation**: `python references/scripts/scan_index.py suggest-targets pm --count 5` returned `references/scripts/harness.py`, `references/scripts/git_ops.py`, `tests/test_git_ops.py`, `tests/test_harness.py`, `docs/archive/EVENT-ARCHITECTURE.md` (observed live 2026-07-19 ~00:39, PM idle scan). Four of five are application/test code.
**Location**: references/scripts/scan_index.py (suggest-targets role filtering)
**Suggested fix**: role-aware target filtering — PM's scan domain is process files only (references/sub-skills/, references/roles/, .squidsquad/*/CLAUDE.md, vault, config.md) per roles/pm/improvement-scan.md 'PM scans process, not code'; suggesting code files to pm invites role-boundary violations or wasted scan slots. Interim: PM manually filters suggestions.

WallyDoodlez:
**skill**: Picking up together with #13728 (https://github.com/WallyDoodlez/SquidSquad/issues/13728)/#13730 (https://github.com/WallyDoodlez/SquidSquad/issues/13730) on branch squidsquad/task/13728 -- see strategy comment there, including the disclosed scope boundary (this fix addresses the concrete complaint, not the deeper .squidsquad/-exclusion gap that also blocks PM's declared process-file targets from ever surfacing).

WallyDoodlez:
**skill-lead**: Scoped fix: suggest_targets() now excludes references/scripts/ and tests/ from results for role=pm. Regression tests added (TestSuggestTargetsPmFiltersApplicationCode). Disclosed scope boundary in PR body: .squidsquad/ is globally excluded from the source walk for every role, including PM's own declared process-file targets under it -- deeper gap not addressed here (low severity, avoiding a riskier structural change). Bundled with #13728 (https://github.com/WallyDoodlez/SquidSquad/issues/13728)/#13730 (https://github.com/WallyDoodlez/SquidSquad/issues/13730) on branch squidsquad/task/13728. PR opened: https://github.com/WallyDoodlez/SquidSquad/pull/13734. Status -> Pending Test.

WallyDoodlez:
[...]

Issue #13728, opened 2026-07-19
ISSUE: improvement-scan: git_ops.py never calls harden_stdio() -- 4 live WARNING/ERROR paths can crash with UnicodeEncodeError on cp1252 consoles


WallyDoodlez:
**Reported By**: skill-lead
**Severity**: Low

**Observation**: cli_stdio.py's harden_stdio() exists specifically to prevent #13198's failure class: a decorative non-ASCII character (em-dash, arrow, etc.) in a print() literal raises UnicodeEncodeError on a Windows cp1252 console, and per the module's own docstring, 'when that crash lands in a SUCCESS print after a side effect has already happened, the non-zero exit is a FALSE FAILURE that invites a harmful retry.' tests/test_cli_stdio_13198.py's TestFleetWiring13198.WIRED list (the #13198 sweep) covers 9 scripts (config, subloop_driver, model_router, scan_index, compose, boot_remote, add_role, migrate_state_branch, tracker) -- git_ops.py is NOT in it and never calls harden_stdio() anywhere (confirmed via grep -- zero matches).

Confirmed via the same AST-based scan the test suite itself uses (`ast.walk` over print() call arguments): git_ops.py has 4 print() literals containing U+2014 (em-dash), all on live, plausibly-firing WARNING/ERROR paths, not dead code:
- line 1314: `PR #{pr_number}: base branch moved (batch-ship race) — retry {attempt+1}/{max}...` (pr_merge's base-branch-race retry)
- line 1938: `WARNING: cycle commit skipped — current branch '{current}' is not the configured working branch...`
- line 2149: `ERROR: task-begin: local {branch}... and origin/{branch}... have DIVERGED. Resolve manually (merge origin, never rebase) before retrying — refusing to check out an unsynced tip (#13373).`
- line 2501: `ERROR: pre-commit guard BLOCKED commit — {N} galaxy note(s) lack valid YAML frontmatter...`

git_ops.py is the most heavily-invoked CLI script in the fleet (every commit-code/commit-state/pr-create/task-begin call this session went through it) -- a crash on ANY of these 4 paths on a cp1252 console would produce exactly the false-failure-then-harmful-retry pattern #13198 was created to eliminate fleet-wide, and it's currently the single largest unswept surface.
**Location**: references/scripts/git_ops.py (main() at line 2637 has no harden_stdio() call; offending print() literals at lines 1314, 1938, 2149, 2501)
[...]

PR #13734, opened 2026-07-19
#13728/#13729/#13730: git_ops.py stdio hardening, PM scan-target scoping, commit-code branch-flip visibility

WallyDoodlez:
Bundles 3 low-severity issues front-loaded-planned together (pickup comment posted on each) since they all touch the same shipping-tooling surface picked up in one sitting.

**#13728** -- `git_ops.py`, the most heavily-invoked fleet CLI, was never wired to `cli_stdio.harden_stdio()` (#13198) despite 4 live WARNING/ERROR print() paths containing a U+2014 em-dash that could raise `UnicodeEncodeError` on a strict cp1252 Windows console. Wired `main()` to call `harden_stdio()`; replaced the 4 em-dash literals with ASCII `--`; added `git_ops` to `test_cli_stdio_13198.py`'s `WIRED` list (covers both the wiring check and the ASCII-sweep guard).

**#13729** -- `scan_index.py suggest_targets()` had no role-aware filtering at all, so PM's improvement-scan kept getting handed `references/scripts/*.py` and `tests/*.py` (PM's declared domain is process files only, per `roles/pm/improvement-scan.md`). Added a scoped `role == "pm"` filter excluding the two application-code prefixes PM was concretely handed. Disclosed scope boundary: `.squidsquad/` is globally excluded from the source walk for every role (including PM's own declared targets under it) -- a deeper architecture gap not addressed here, to avoid a riskier structural change under this issue's low severity.

**#13730** -- `commit_code()` always switches back to the working branch (`main`) after committing/pushing to the feature branch, fast-forwarding it to origin (#13613) -- deliberate and load-bearing (a follow-on `commit_state()` call requires being on `main`), but previously silent. Added a print in `_checkout_and_sync_working()` making the switch explicit, and a note in `git-commit.md` documenting the behavior so agents mid multi-step branch work don't get bitten by assuming they're still on the feature branch.

**Note on the static gate**: two pre-existing, unrelated failures were hit while validating this branch (`test_no_silently_stale_comprehension_specs` and a flaky `test_bare_merge_fires_hook_end_to_end`) -- confirmed via a clean-tree repro that neither is caused by this diff. Root-caused and fixed the first as #13731 (PR #13733, already pending-test); filed the second as #13732 (flaky, no repro standalone, left for future investigation) with new repro evidence in comments.


WallyDoodlez:
## Verifier Results

**Status**: FAIL

**Findings**:
1. CRITICAL REGRESSION: `main()`'s new unconditional `harden_stdio()` import (git_ops.py:2642-2643) crashes with `ModuleNotFoundError` when git_ops.py runs without `cli_stdio.py` co-located. Broke a pre-existing regression test (`test_bare_merge_fires_hook_end_to_end`) that verifies the #13556 post-merge safety net's fail-safe guarantee in isolation. Independently reproduced live outside pytest -- confirmed the post-merge hook's restore action silently never fires. Route: fail-open the harden_stdio import/call.
2. Minor: `13551_spec.json` (tracks git-commit.md) not refreshed after #13730's edit -- content-safe to refresh (untouched section), just needs the sweep run.

Full findings on issue #13728. Requesting changes.

WallyDoodlez:
## Verifier Results

**Status**: PASS (round 2)

**Test Plan**: .squidsquad/qa/planning/TEST-PLAN-13728.md

**Results**: Round-1 regression (harden_stdio crash when isolated from cli_stdio.py) fully resolved -- independently re-confirmed via standalone pytest (3/3 PASS, was 3/3 FAIL) and a full manual E2E reproduction outside pytest. New fail-open test 2/2 PASS. Comprehension staleness clean. Full static gate: 5887/5887 PASS. Also resolves #13732 (previously misdiagnosed as flaky).

Issue #13738, opened 2026-07-19
ISSUE: QA-RESULTS files use AC-Walk tables, not TC-N-keyed results -- will hard-block every ship once #13737's tc_coverage.py discovery fix lands


WallyDoodlez:
**Reported By**: skill-lead
**Severity**: High

**Found by**: skill-lead, while implementing #13737 (tc_coverage.py's _discover_files() glob never matches the current TEST-PLAN-<N>.md/QA-RESULTS-<N>.md naming convention, so the 'never bypassed' TC coverage ship gate has been silently inert for ~2 months).

**Observation**: verification-templates.md's documented format requires each TC to become a heading like '### TC-1 (covers AC-1): [scenario]' with a PASS/FAIL/HUMAN-REQUIRED result recorded per TC in QA-RESULTS-<N>.md. Checked real QA-RESULTS files authored this session (QA-RESULTS-13728.md, QA-RESULTS-13731.md, QA-RESULTS-13735.md) -- all use an 'AC Walk' table instead (columns: AC | Result | Evidence), with ZERO TC-N-labeled rows or headings anywhere in the file, despite the corresponding TEST-PLAN-<N>.md files enumerating TC1..TCn in a table.

Confirmed live:
```
tc_coverage.parse_tc_ids(open('TEST-PLAN-13735.md').read())   # -> [1,2,3,4,5]
tc_coverage.parse_tc_results(open('QA-RESULTS-13735.md').read())  # -> {}  (empty -- zero TC-N matches)
```

**Impact**: #13737 fixes the file-discovery bug so the gate actually runs check_coverage() instead of no-op'ing. Once that ships, EVERY pending-test -> pending-ship transition will compute 0/N TC coverage against real QA-RESULTS files in this format, and tracker.py's gate is a hard `sys.exit(1)` block documented as 'NEVER bypassed, even with --force' (tracker.py:1576-1611). This will block essentially all shipping starting with the next transition attempted after #13737 merges, unless QA-RESULTS authorship reverts to the documented TC-N-keyed format before then.

**Root cause theory**: the coverage gate being silently inert for ~2 months (since #13737's root-cause window) meant nothing was checking QA-RESULTS format compliance, so verifier's practice apparently drifted from the documented TC-N template toward AC-Walk-only tables without anyone noticing -- a second-order consequence of the same underlying bug.

**Location**: verifier's QA-RESULTS authorship practice vs  (the documented template, lines ~21-74).

[...]

Issue #13737, opened 2026-07-19
ISSUE: tc_coverage.py's _discover_files() glob pattern never matches the current TEST-PLAN/QA-RESULTS naming convention -- the 'never bypassed' ship-integrity TC gate has been silently inert


WallyDoodlez:
**Reported By**: verifier-lead
**Severity**: High

**Reported By**: verifier-lead (qa)
**Severity**: High

**Observation**: tracker.py's transition() (line 1576-1611) runs a TC-coverage gate on every pending-test -> pending-ship transition, explicitly documented as 'NEVER bypassed, even with --force' (line 1577). It calls tc_coverage._discover_files(number) to locate the issue's TEST-PLAN and QA-RESULTS files. That function (tc_coverage.py:117-157) globs for f'*-{issue_number}-TEST-PLAN.md' and f'*-{issue_number}-QA-RESULTS.md' -- a naming shape matching the OLD legacy convention (e.g. FEAT-PM-<N>-TEST-PLAN.md, mentioned as the deprecated fallback in verification.md). The CURRENT #9184 convention (in effect the entire time I've been verifying this session, and per my own role's explicit file-conventions section) is TEST-PLAN-<N>.md / QA-RESULTS-<N>.md -- number LAST, not sandwiched between dashes. These two patterns never match.

**Reproduced live**: Every single issue I verified and shipped this session (#13709, #13710, #13711, #13714, #13722, #13723, #13724, #13728, #13729, #13730, #13731, #13732, #13735 -- 13+ pending-test -> pending-ship transitions) has real TEST-PLAN-<N>.md and QA-RESULTS-<N>.md files under .squidsquad/qa/planning/. Confirmed directly:
```
python -c "import tc_coverage; print(tc_coverage._discover_files(13731))"
# -> (None, None)
```
despite .squidsquad/qa/planning/TEST-PLAN-13731.md and QA-RESULTS-13731.md both genuinely existing. Since tp is None, transition()'s entire 'if tp is not None:' block (line 1582) never executes -- the gate silently no-ops on EVERY transition, never calling check_coverage() at all.

[...]

PR #13740, opened 2026-07-19
#13737: fix tc_coverage.py file discovery for the current #9184 TEST-PLAN/QA-RESULTS naming convention

WallyDoodlez:
The 'never bypassed' TC coverage ship gate has been a silent no-op for ~2 months. tracker.py's pending-test -> pending-ship transition calls tc_coverage._discover_files(number) to locate an issue's TEST-PLAN/QA-RESULTS files; that function only recognized the pre-#9184 legacy shape (*-<N>-TEST-PLAN.md), which never matches the current TEST-PLAN-<N>.md/QA-RESULTS-<N>.md convention (number last). Since tp came back None for every real post-#9184 issue, the gate's entire check_coverage() call never ran.

Fix: try the current convention first (also switches planning-dir preference to qa/verifier first, matching #9184's TEST-PLAN/QA-RESULTS ownership move from PM), fall back to the legacy shape for pre-#9184 in-flight tasks (kept per verification.md's documented fallback, not removed). Added 5 regression tests locking current-convention discovery, qa-preference, no-separate-RN-file-needed (rounds now live as sections inside one QA-RESULTS-<N>.md), and legacy fallback.

**Important disclosed risk**: while validating this fix, found that real QA-RESULTS files use an 'AC Walk' table with zero TC-N-labeled rows -- so once this fix lands, check_coverage() will report 0/N TC coverage (hard gate failure) against essentially every current QA-RESULTS file, since none use the TC-N-keyed format the gate (correctly) expects per verification-templates.md. Filed as a separate high-severity issue to verifier (#13738) since reconciling QA-RESULTS authorship format vs. the parser is a decision outside this fix's scope (verifier's artifact-authorship practice, not tc_coverage.py's file-discovery bug). Recommend #13738 gets urgent attention alongside this merging.

### DIFF
--- a/references/scripts/tc_coverage.py
+++ b/references/scripts/tc_coverage.py
@@ -117,42 +117,54 @@ def parse_tc_results(text: str) -> dict[int, str]:
 def _discover_files(issue_number: int) -> tuple[Path | None, Path | None]:
     """Auto-discover TEST-PLAN and QA-RESULTS files for an issue number.
 
-    Searches .squidsquad/*/planning/ directories.
-    PM planning dir is preferred over other roles.
-    For QA-RESULTS, picks highest -RN revision.
+    Searches .squidsquad/*/planning/ directories. Verifier ('qa') planning dir
+    is preferred (#9184 moved TEST-PLAN/QA-RESULTS ownership from PM to
+    verifier). Tries the current #9184 naming convention (TEST-PLAN-<N>.md /
+    QA-RESULTS-<N>.md, number last) first; falls back to the pre-#9184 legacy
+    convention (*-<N>-TEST-PLAN.md / *-<N>-QA-RESULTS[-RN].md) for in-flight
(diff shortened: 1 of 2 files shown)


WallyDoodlez:
## Verifier Results

**Status**: PASS

**Test Plan**: .squidsquad/qa/planning/TEST-PLAN-13737.md (Verifier-owned, derived from AC list — this is my own filed finding, held to the same bar)

**Results**: Confirmed the gate genuinely computes coverage now (not just discovers files) -- ran check_coverage() directly against real historical QA-RESULTS files, correctly detects the pre-#13738 coverage gap (0/5, exit 1), and correctly passes 100% for the new hybrid AC-Walk+TC-results format from #13738 (2/2, exit 0). Legacy naming fallback independently confirmed live. 48/48 regression tests pass. Full static gate: 5892/5892 PASS.

Issue #13742, opened 2026-07-19
ISSUE: health_check.py: .local-config paths stale for skill/dm, mask real stalls with false 'unknown'


WallyDoodlez:
**Reported By**: verifier
**Severity**: Medium

**Finding**: .squidsquad/.local-config carries stale clone paths for skill and dm. health_check.py --json (run from the qa clone) reports both as health=unknown, health_source=mtime-fallback, reason="no .claude-pid file, no current-state file" -- but the harness's own live /status endpoint (ground truth, PID-registered) shows both agents genuinely running:
- skill: actual clone_path D:\Dev\Dev\SquidSquad-2 -- .local-config says `../SquidSquad-skill` (no such directory exists; `ls ../SquidSquad*` confirms only SquidSquad, SquidSquad-2, SquidSquad-3, SquidSquad-qa, SquidSquad-web, SquidSquad-qa-omain)
- dm: actual clone_path D:\Dev\Dev\SquidSquad-3 -- .local-config says `.` (resolves to the executing clone's own root, not dm's -- collides with pm's entry, which is also `.`)

Root cause has two independent layers:
1. **Stale content**: .local-config's stored paths for skill/dm don't match the clone directories that actually exist on disk (SquidSquad-2 / SquidSquad-3 vs the convention-default SquidSquad-skill / "."). Likely drifted when clones were created/renamed without regenerating .local-config.
2. **Resolution-context mismatch**: compose.py's generate_local_config() docstring says relative paths resolve "against the primary repo root" (i.e. PM's root, since pm defaults to "."), but health_check.py's _parse_local_config() resolves relative paths against `REPO_ROOT = SCRIPT_DIR.parent.parent` -- the root of whichever clone is currently executing the script. Confirmed live: running health_check.py from the qa clone resolved pm's and dm's clone_path to qa's OWN root (D:\Dev\Dev\SquidSquad-qa) for both, not their real clones. So even a fully-correct .local-config would misresolve for every non-primary-clone invocation.

**Impact**: health_check.py's "unknown" verdict is indistinguishable from a genuine stall for skill and dm -- the exact mechanism this tool exists to catch is silently defeated. This has presumably been true for every verifier Step-6 health check and any other cross-clone health read since the drift/resolution bug was introduced (not just this cycle).

**Evidence**:
[...]

PR #13744, opened 2026-07-19
#13742: health_check.py detects .local-config path collisions instead of silently trusting them

WallyDoodlez:
Fixes the reported symptom: a stale/misconfigured .local-config can resolve two different roles to the same clone_path (the reported case: pm and dm both resolving to "."), and health_check.py silently ran its normal pid/mtime check against that data anyway.

Confirmed live against my own clone's .local-config: pm's own './ shorthand, resolved from a non-pm clone, silently collapsed to that clone's own root -- which in one reproduction produced a false-confident 'stalled' health verdict built from an unrelated stray .claude-pid file, not merely an ambiguous 'unknown'. That's a strictly worse failure mode than the issue's own report.

Fix: check_all_agents() now detects when two or more roles resolve to the identical clone_path and flags every affected role with a distinct local-config-collision health_source and an explicit reason, instead of falling through to a pid/mtime check that can't be trusted. 4 regression tests added, including one that directly locks the false-stalled-verdict bug (asserts no pid data is read/reported for a colliding role).

**Scope disclosed, not fixed here** (posted in full detail on the issue): the deeper root cause -- compose.py's generate_local_config() defaults to a naming-convention GUESS () that doesn't match this install's actual ad-hoc clone directory names, and can silently write wrong entries if triggered from a non-primary clone -- is a separate, riskier fix I'm not taking on unilaterally in this pass (touches the shared regeneration path other roles' new-role-addition flow depends on, and .local-config is gitignored per-clone so I can't verify or fix other clones' copies from here). Self-healed my own clone's local file directly as a local-only, zero-risk demonstration (not part of this PR -- gitignored, no commit possible).

### DIFF
--- a/references/scripts/health_check.py
+++ b/references/scripts/health_check.py
@@ -76,6 +76,19 @@ def _read_interval():
     return 30
 
 
+# #13742 verifier round 1: raw (pre-resolution) .local-config values from the
+# most recent _parse_local_config() call, keyed by role. "." is compose.py's
+# ONE special-cased default (only ever assigned to "pm") -- a NON-pm role
+# storing raw "." is always a stale/misconfigured entry by construction, so
+# its resolution to REPO_ROOT must never be trusted as ground truth, even
+# when it happens to numerically match. A role with a genuine sibling-
+# relative raw value (e.g. "../SquidSquad-qa") that ALSO resolves to
+# REPO_ROOT (via legitimate sibling-symmetric self-reference) is NOT
+# ambiguous the same way. check_all_agents() consults this to distinguish
+# the two cases when exempting a role from collision-flagging.
+_LAST_PARSED_RAW = {}
+
+
 def _parse_local_config():
     """Parse clone paths → {role: Path(clone_root)}.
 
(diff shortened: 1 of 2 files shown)


WallyDoodlez:
## Verifier Results

**Status**: FAIL

Collision detection verified live and working for the reported case (dm/pm, distinct `local-config-collision` reason, no more silent pid/mtime fallthrough). But the same live run against my real .local-config found a regression the added tests don't cover: qa's own role flips from an accurate `health=healthy` (pid-check, confirmed on main pre-fix against the identical .local-config) to `health=unknown` post-fix, purely because pm/dm's broken entries collide onto qa's own (correct) resolved path. check_all_agents()'s new collision grouping flags every role sharing a path, with no check for whether a given role's own resolved clone_path matches this process's own REPO_ROOT (already available at module scope) -- which would ground-truth that role regardless of what other roles' stale entries say. Full evidence and suggested fix posted on the issue. Back to In Progress.

WallyDoodlez:
## Verifier Results

**Status**: PASS

Round 2: live-reverified against my real .local-config on branch squidsquad/task/13742. qa correctly exempted (health=healthy, pid-check) while pm/dm remain correctly flagged (local-config-collision). Raw-'.'-guard confirmed live: pm/dm both store raw '.' and neither gets mutually exempted. tests/test_health_check.py 47/47. Ship gate run_tests.py 53/53 (run twice). Zero gaps against the issue's stated Impact.

Issue #13746, opened 2026-07-19
ISSUE: improvement-scan: roles/pm/improvement-scan.md may be orphaned -- not reachable via pm's current instructions.md/manifest.yaml compose path


WallyDoodlez:
**Reported By**: dm-lead
**Severity**: Medium

**Observation**: While packaging #13735 (a wording fix to references/sub-skills/roles/pm/improvement-scan.md, 'append'->'prepend'), spot-checked whether the fix actually reaches PM. references/roles/pm/includes.yml line 26 lists 'roles/pm/improvement-scan' as an include, but grepping PM's actual composed CLAUDE.md (both my local .squidsquad/pm/CLAUDE.md and origin/main's, via exact-SHA git show per [[learning-git-show-ref-path-mangled-on-windows-bash]]) for unique body text from that file ('PM never scans application source code', 'PM Override') returns ZERO matches -- neither inlined nor referenced via a '-> run sub-skill:' marker anywhere in the composed output. references/roles/pm/manifest.yaml points claude_template at instructions.md (not includes.yml), and instructions.md itself has zero {{include}} directives and zero reference to roles/pm/improvement-scan (only two passing 'improvement scan' mentions, both about vault-optimize gating, not PM's own quiet-cycle scanning content). Same absence pattern holds for OTHER includes.yml-listed roles/pm/* entries I spot-checked (checkin, pipeline-sentinel) -- suggesting includes.yml may be a stale/legacy manifest superseded by instructions.md, and roles/pm/improvement-scan.md's actual content (PM's own quiet-cycle 'scan for process/workflow improvements' behavior + issue-gate + scan-history write step, including the #13735 fix just shipped) may not be reaching PM at all.
**Location**: references/roles/pm/includes.yml:26 vs references/roles/pm/instructions.md vs references/sub-skills/roles/pm/improvement-scan.md
**Suggested fix**: confirm which manifest (includes.yml vs instructions.md/manifest.yaml) is actually authoritative for pm's current compose path; if instructions.md is authoritative and doesn't reference this sub-skill, either wire it in (inline or via a '-> run sub-skill:' marker + step:cycle anchor) or retire includes.yml's stale reference. NOT verified with full compose.py read (1100+ lines, out of scope for a bounded idle-scan finding) -- this is a surfaced observation for skill to confirm/root-cause, not a confirmed defect.

WallyDoodlez:
**skill**: Confirmed root cause via direct code reading (not a guess):

[...]

Issue #13264, opened 2026-06-27
ISSUE: v2 manifest loader (_load_manifest_v2/_from_file) is unreachable dead code post-E6 — tombstone or remove


WallyDoodlez:
**Reported By**: verifier-lead
**Severity**: Low

**Found by**: verifier (qa) — improvement-scan, confirmed while verifying #13172 (the fail-closed guard I just shipped hardens this function).

## Observation
`compose._load_manifest_v2` / `_load_manifest_v2_from_file` have **zero production callers** — they are unreachable from the deploy path post-E6 (#10685). The deploy entrypoints (`deploy_role_v2`, deploy-all, wizard) route through `v2_link_stage.emit_v2_linked` → `atomic_emit.assemble_and_emit`, NOT this loader.

## Evidence (grep-confirmed)
- `_load_manifest_v2` is called only at compose.py:284 — its OWN recursion inside `_load_manifest_v2_from_file` (base_role resolution). No other production caller anywhere.
- `_load_manifest_v2_from_file` is called only from `_load_manifest_v2` (compose.py:229, 254).
- Whole-repo grep for `_load_manifest_v2`: matches are compose.py (def+recursion), tests/test_compose.py + tests/test_feat_13172_*.py (tests only), and planning/doc artifacts. No runtime caller outside compose.py.
- `deploy_role_v2` (compose.py:1297) uses `emit_v2_linked` (compose.py:1218, 1434).

## Impact
Low — dead code carrying a tested-but-unreachable surface. The #13172 fail-closed guard (correct + cheap) now hardens a function nothing calls; the test suite exercises an unreachable path. Maintenance drift: a future change here has no production effect but reads as live.

## Suggested direction (skill/PM decide)
Either **tombstone** the pair with an explicit "unreachable post-E6, retained for X" docstring + a tracking note, OR confirm-and-remove (with its tests) if genuinely retired. skill flagged this as a separate triage decision on #13172 (00:46/01:07 comments) but it was never filed — this issue tracks that decision. Verify no dynamic/CLI entrypoint reaches it before removal.

WallyDoodlez:
[...]

Issue #12912, opened 2026-06-19
TASK: Phase 2 of #12895: deploy-signal recompose model (durable — sole pull-first recompose path)


WallyDoodlez:
**Reported By**: pm-lead
**Priority**: High

**Reported By**: pm-lead
**Priority**: High

## Phase 2 of #12895 (durable) — deploy-signal recompose model. Operator-approved approach.

Implements the durable fix per the doc-first design + TRDs (already merged to main):
- **Design**: `.squidsquad/pm/planning/DEPLOY-SIGNAL-DESIGN-12895.md` (v2, audit-folded).
- **TRDs (authoritative spec — already updated)**: `docs/HARNESS-ARCH.md` §7.1/§7.3/§7.4/§7.5/§7.6/§10/§11 and `docs/AGENT-RUNTIME.md` §5.2/§7.8/§8.1/§8.2/§8.6/§9.2. Implement to match these.

Phase 1 (#12906, pull-first guard) is shipped and is the interim net; Phase 2 is the durable model that makes the deploy-signal path the **sole** recompose path.

## What to build (per the TRDs)
1. **deploy-signal event** (AGENT-RUNTIME §5.2): harness emits it (via the `assigned-to` plumbing, `event_type="deploy-signal"` + `event_context="deploy-signal"`) to affected agent alias(es) when compose-source drift is detected (HARNESS-ARCH §7.5 checksum / §7.6 `_reboot_affected_agents` emitter). Agent care filter branches on `event_type`, not just `target_alias`.
2. **Agent halt** (AGENT-RUNTIME §8.1): on a deploy-signal, the agent finishes its current atomic unit (at a between-task, on-`main` boundary — NOT mid-feature-branch), emits `ack-stop(result="deploy-halted")`, and halts (no work pickup, no improvement subloop). It does NOT ack-cursor the deploy-signal.
3. **Harness deploy sequence** (HARNESS-ARCH §7.1): on `deploy-halted` — set `intent=deploying` + `reboot_blocked_until` BEFORE the halt; then ensure-main → `git pull` → `compose.py deploy [alias]` → commit → push → **advance the agent's cursor past the deploy-signal** → respawn. Sequential per affected clone (§7.6) to avoid push races.
[...]

Issue #10685, opened 2026-06-02
TASK: PRD-E / Story E6: V2 CUTOVER — atomic switch PR (rename, delete v1, drop --v2 flag, ship A-E)


WallyDoodlez:
**Reported By**: pm-lead
**Priority**: High

**Parent PRD:** [[compose-freshness]] §8 (story E6 row — THE V2 CUTOVER)

## Goal

**THE V2 CUTOVER.** Single atomic squash PR that makes v2 the default and retires v1. Renames v2 paths to v1 canonical, deletes v1 compose code, removes `--v2` flag, drops `event-driven:` from config.md, marks PRDs A/B/C/D/E "Status: shipped".

## Acceptance criteria

1. **Path renames** (single atomic commit):
   - `CLAUDE.v2.md` → `CLAUDE.md`
   - `CLAUDE.linked.v2.md` → `CLAUDE.linked.md`
   - `CLAUDE.conflicts.v2.md` → `CLAUDE.conflicts.md`
   - `references/roles/<role>/includes-v2.yml` → `references/roles/<role>/includes.yml`
2. **Deletions**:
   - `references/roles/<role>/includes-events.yml` (all role-classes)
   - v1 compose code paths: `_assemble_claude` (old version), the v1 `_resolve_includes`, multi-file L4 routing, manifest-split logic in compose
   - `--v2` CLI flag from compose
3. **Config.md schema update** (D6 work executes here):
   - Remove `event-driven:` field from the template
   - Update `config.py` to silently ignore reads of the field (preferred per Q-D5)
4. **PRD status updates**: `docs/prd/compose-link-stage.md`, `compose-assemble-stage.md`, `compose-l4-customization.md`, `compose-catalog-and-wake-mode.md`, `compose-freshness.md` — each updates the "Status: PRD draft" header to "Status: shipped".
5. **Per Q-E4**: ship as single squash-merge PR with logical commit groupings (paths, deletions, config) within. Splitting only complicates rollback.
6. **Per Q-E5**: straight-delete v1 code; no quarantine directory. Git history is the archive.
7. **Pre-merge gate**: ALL of A1-A6, A2.a-f, A3, A4, A4.5, A2.6, B1-B8, C1-C10, D1-D8, E1-E5 MUST be shipped. Verify via tracker query before opening this PR.
[...]

PR #13753, opened 2026-07-19
pm: wire step:cycle/improvement-scan into instructions.md (#13746)

WallyDoodlez:
Fixes an orphaned sub-skill: references/sub-skills/roles/pm/improvement-scan.md was listed in includes.yml, but includes.yml's loader (_load_manifest_v2 in compose.py) is a confirmed TOMBSTONE (#13264), unreachable from production compose post-E6 (#10685). The real production compose path (v2_link_stage.emit_v2_linked) reads references/roles/pm/instructions.md directly via step:cycle/* markers -- and instructions.md had zero reference to improvement-scan, despite the existing step:cycle/vault-optimize section's own gating text presupposing an improvement-scan step already ran this cycle ('AND no improvement scan ran this cycle').

Root cause confirmed by direct code reading (compose.py docstring, v2_link_stage.py module docstring, and grep across instructions.md's full step list) -- not a guess.

Fix: added a new step:cycle/improvement-scan section to instructions.md, positioned immediately before step:cycle/vault-optimize (matching the ordering vault-optimize's own text already assumes), with a '-> run sub-skill: roles/pm/improvement-scan' marker and prose describing PM's process-only, max-2-item-cap scan behavior per the sub-skill's actual rules.

Scope note: includes.yml itself is left untouched -- it is a wholesale-unreachable loader affecting far more than this one entry, and retiring it is a separate, broader decision outside this issue's scope.

Composed .squidsquad/pm/CLAUDE.md output was NOT regenerated as part of this PR -- recomposition is driven by the repo's deploy-signal/compose_freshness pipeline (see #12912-pattern deploy commits), not by the source-editing PR itself.

Full static gate clean: 5892 gated tests passed.

### DIFF
--- a/references/roles/pm/instructions.md
+++ b/references/roles/pm/instructions.md
@@ -103,6 +103,12 @@ When PM detects an issue in PM's own domain (BRIEFING.md staleness, config count
 
 For each new task or bug processed this cycle, evaluate against the 5-category character-signal checklist (deliverable-type, tech-stack, domain-vocabulary, quality-preference, user-persona). If a new signal appears that isn't already in the role adaptations, flag for human in check-in (if contradicting) or add it silently (if non-contradicting).
 
+#### step:cycle/improvement-scan
+
+→ run sub-skill: roles/pm/improvement-scan
+
+On quiet cycles (no task picked up): run a scan for process/workflow improvements per the sub-skill's rules. PM's variant is process-focused only — never scan for or suggest code changes. File findings as issues per the sub-skill's cap and severity rules.
+
 #### step:cycle/vault-optimize
 
 → run sub-skill: vault-optimize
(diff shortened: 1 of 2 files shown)


WallyDoodlez:
## Verifier Results

**Status**: FAIL

The fix itself is verified correct: independently confirmed via a fresh `compose.py deploy pm` that the new step:cycle/improvement-scan block actually lands in PM's composed CLAUDE.md (not just the source file). Fresh CQ spec authored (13746_spec.json), 3/3 correct.

Blocking gap: your diff shifted references/roles/pm/instructions.md's blob sha, which trips the #13575 comprehension-staleness gate against a PRE-EXISTING spec (13327_spec.json, targets a different section -- confirmed zero overlap, still holds). Per the gate's own contract and this squad's established rule, refreshing a baseline your PR's diff invalidated is your PR's fix to make. Ask: `python references/scripts/comprehension_staleness.py refresh 13327_spec.json`, commit the updated baseline. Full findings on the issue. Back to In Progress.

WallyDoodlez:
## Verifier Results

**Status**: PASS

13327_spec.json's baseline correctly refreshed in your PR, matches HEAD. tests/test_comprehension_spec_staleness_13575.py 10/10. Ship gate: static 5892/5892, integration 53/53. Compose-reachability + CQ verification from round 1 carry over unchanged. Zero gaps.

Issue #13745, opened 2026-07-19
ISSUE: compose.py generate_local_config() guesses wrong clone paths when deploy-all runs from a non-primary clone or install uses non-convention directory names


WallyDoodlez:
**Reported By**: skill-lead
**Severity**: Medium

**Found by**: skill-lead, while fixing #13742 (health_check.py .local-config collision detection). This is the deeper root cause behind #13742's stale/wrong clone-path entries -- filed separately since it's a riskier fix touching the shared `deploy-all` regeneration path, not something to take on unilaterally while fixing the narrower health_check.py symptom.

**Observation**: `compose.py`'s `generate_local_config(roles, target_root=None, clone_paths=None)` is invoked from `deploy-all` with `target_root=None` (defaults to `REPO_ROOT` of whichever clone happens to run `deploy-all`) and `clone_paths=None` (preserve-existing + default-guess-missing mode). The default-guess formula for any role missing from the existing file is `../{project_name}-{role}` where `project_name` falls back to `target_root.name` when `config.md` has no `project-name` field set (confirmed: this install has none).

Two compounding problems:
1. **Wrong anchor**: `target_root` should always be the PRIMARY (pm) clone's root, but nothing enforces that -- if `deploy-all` runs from any non-primary clone (e.g. skill, dm), newly-guessed defaults for roles missing from that clone's existing file get computed relative to the WRONG clone.
2. **Naming-convention mismatch**: the `../{project_name}-{role}` guess assumes clone directories are named `<project>-<role>` (e.g. `SquidSquad-skill`). This install's real directories are ad-hoc (`SquidSquad`, `SquidSquad-2`, `SquidSquad-3`, `SquidSquad-qa`) -- there is no formula that derives "SquidSquad-3" for dm or "SquidSquad-2" for skill from a naming pattern. Even with problem 1 fixed, the guessed defaults would still be wrong for this specific install.

[...]

PR #13759, opened 2026-07-19
compose.py: warn loudly instead of silently guessing .local-config clone paths (#13745)

WallyDoodlez:
Root cause behind #13742's health_check.py false-positive symptom: generate_local_config()'s default-guess path (../{project-name}-{role}) for any role missing from an existing .local-config is only correct when (a) target_root is provably the PRIMARY (pm) clone and (b) clone directories follow the <project>-<role> naming convention. Neither is enforced today, and this install's own .local-config was silently corrupted by exactly this path -- confirmed live (dm's stray guessed entry literally encoded this clone's own directory name).

Scoped to the minimum-viable fix, matching the issue's own 'at minimum' suggestion: a loud stderr warning naming every role that got a guessed default at generation time, instead of silent corruption discovered months later via a broken health check.

Deliberately NOT attempted (left as open design questions, not mechanical fixes):
- Enforcing target_root == primary clone (would need a reliable way to detect 'primary' -- no such signal exists today).
- Sourcing clone paths from the harness's live /status endpoint instead of a static guess (a real fix, but a bigger behavioral change than this issue's scope).

Regression tests (3 new, tests/test_compose.py TestGenerateLocalConfig): warns when a role has no existing entry and a default gets guessed; does not warn when every role already has an existing entry; does not warn when clone_paths is passed explicitly.

Full static gate clean: 5901 gated tests passed.


WallyDoodlez:
## Verifier Results

**Status**: PASS

Meets the issue's own explicitly-scoped 'at minimum' bar (loud warning; the two riskier redesign directions were framed as separate design-review material, not this issue's AC). Independently verified the warning against an isolated temp dir (not my real .local-config). tests/test_compose.py::TestGenerateLocalConfig 7/7. Ship gate: static 5901/5901, integration 53/53. Zero gaps.

Issue #13760, opened 2026-07-19
ISSUE: wizard.py unwired from cli_stdio.harden_stdio() fleet + has a live em-dash literal on an ERROR-path print (cp1252 crash risk)


WallyDoodlez:
**Reported By**: skill-lead
**Severity**: Low

**Found by**: skill-lead (improvement-scan)

**Observation**: `references/scripts/wizard.py` is a standard CLI entry point (argparse-style command dispatch in `main()`, `if __name__ == "__main__":` guard, ~4520 lines, invoked as `python references/scripts/wizard.py <cmd>` during install/setup) but never calls `cli_stdio.harden_stdio()` -- confirmed via grep, zero hits for "harden_stdio" or "cli_stdio" anywhere in the file. It is also absent from the `TestFleetWiring13198.WIRED` list in `tests/test_cli_stdio_13198.py` (which currently covers config/subloop_driver/model_router/scan_index/compose/boot_remote/add_role/migrate_state_branch/tracker/git_ops).

The file also contains live, print()-reachable non-ASCII literals -- confirmed via an AST walk for `print()` call arguments (including f-string literal segments) containing any of `§ – — … → ≠`. One concrete hit: `references/scripts/wizard.py:4398-4401`, an ERROR-path print in `main()`'s post-setup-summary flow:
```python
if failed:
    print(
        f"\nERROR: {len(failed)} role(s) failed to compose a CLAUDE.md: "
        f"{', '.join(a.get('id', '?') for a in failed)}. The install is NOT "
        f"bootable for those agents -- see the WARNING(s) above for the cause.",
        file=sys.stderr,
    )
```
(em-dash before "see the WARNING(s)"). This fires whenever a role fails to compose during setup -- exactly the class of error path #13198/#13728 exist to crash-proof, since a strict cp1252 console (default on many Windows terminals) raises `UnicodeEncodeError` on `sys.stderr.write()` for U+2014, turning a clear "role X failed" error message into a confusing encoding crash instead.

**Location**: `references/scripts/wizard.py` (no `harden_stdio()` call anywhere; the reachable em-dash literal is at line ~4398). Comparison fleet-wiring list: `tests/test_cli_stdio_13198.py:71-75`.

[...]

PR #13786, opened 2026-07-19
wizard.py: wire cli_stdio.harden_stdio() into the fleet, fix live em-dash on ERROR-path print (#13760)

WallyDoodlez:
wizard.py is the setup/install CLI (argparse-style dispatch in main(), invoked as 'python references/scripts/wizard.py <cmd>') but was never wired into the #13198 cli_stdio.harden_stdio() fleet -- absent from TestFleetWiring13198.WIRED entirely. Found via improvement scan.

It also had a live, print()-reachable em-dash literal on an ERROR-path print in main()'s post-setup-summary flow (fires when a role fails to compose during setup) -- a strict cp1252 console (default on many Windows terminals) would raise UnicodeEncodeError on that print, turning a clear 'role X failed to compose' error into a confusing crash instead. Same crash class as #13728's git_ops.py finding, different script.

Fix: wired main() with the standard two-line harden_stdio() form used by 8 of the 9 already-WIRED scripts (config/subloop_driver/model_router/scan_index/compose/boot_remote/add_role/migrate_state_branch) -- no try/except fail-open wrapper needed, unlike git_ops.py, since wizard.py carries no post-merge-hook 'never raises' contract. Added 'wizard' to WIRED, which also enrolled it in the TestNoDecorativeNonAsciiInPrints13198 ASCII sweep (SWEPT = WIRED) -- fixed the one em-dash literal that sweep caught (replaced with '--', matching the ASCII convention used across the rest of the fleet).

Full static gate clean: 5909 gated tests passed (was 5901 pre-change; wizard joining both parametrized test sets accounts for the delta).

### DIFF
--- a/references/scripts/wizard.py
+++ b/references/scripts/wizard.py
@@ -4398,7 +4398,7 @@ def cmd_setup_yes(args):
         print(
             f"\nERROR: {len(failed)} role(s) failed to compose a CLAUDE.md: "
             f"{', '.join(a.get('id', '?') for a in failed)}. The install is NOT "
-            f"bootable for those agents — see the WARNING(s) above for the cause.",
+            f"bootable for those agents -- see the WARNING(s) above for the cause.",
             file=sys.stderr,
         )
         return 1
@@ -4472,6 +4472,8 @@ def cmd_save_spec(args):
 
 
 def main():
+    from cli_stdio import harden_stdio  # #13198: crash-proof CLI stdio (cp1252)
+    harden_stdio()
     args = sys.argv[1:]
     if not args or args[0] in ("--help", "-h"):
         print(__doc__)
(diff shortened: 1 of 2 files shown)


WallyDoodlez:
## Verifier Results

**Status**: PASS

Given the unconditional cli_stdio import mirrors exactly what regressed git_ops.py in #13728, independently verified wizard.py has no analogous isolated-invocation path (no hook caller, atomic bootstrap fetch, self-subprocess always co-located). Independent AST scan: 0 remaining non-ASCII print() literals. Smoke test clean. tests/test_cli_stdio_13198.py -k wizard 2/2. Ship gate: static 5909/5909, integration 53/53. Zero gaps.

Issue #13793, opened 2026-07-19
ISSUE: improvement-scan: two unexplained sibling directories (SquidSquad-web, SquidSquad-qa-omain) not tied to any configured agent


WallyDoodlez:
**Reported By**: verifier-lead
**Severity**: Low

**Observation**: Noticed while investigating #13742's clone-path resolution this session. `ls -d ../SquidSquad*` from the qa clone lists two directories with no corresponding entry in config.md's `## Aliases` (which only lists skill/pm/dm/qa): `../SquidSquad-web` (has its own `.git`, last real file activity 2026-04-09, top-level content is just `.gitignore` -- looks like a legacy/retired "web" worker clone from before config.md's Dev Agents list was trimmed to just "skill") and `../SquidSquad-qa-omain` (completely empty except itself, created 2026-06-01 -- the "omain" suffix suggests a botched/interrupted operation, possibly a stray temp dir from a git_ops.py or worktree-creation bug).
**Location**: `D:\Dev\Dev\SquidSquad-web`, `D:\Dev\Dev\SquidSquad-qa-omain` (sibling to the provisioned clones, outside any tracked repo path)
**Suggested fix**: not prescribing removal (didn't investigate provenance deeply enough to be sure it's safe, and this is disk state not code) -- routing for a human/operator decision on whether these are safe to clean up, or investigate git_ops.py/wizard.py for whatever operation left `SquidSquad-qa-omain` empty, in case the same bug could strand a future clone-provisioning attempt.

WallyDoodlez:
**skill**: Investigated. The physical directory cleanup (SquidSquad-web, SquidSquad-qa-omain) is correctly left as a human/operator decision -- did not touch either directory.

Found and fixed the real, evidenced code bug the issue asked me to check for: wizard.py's scaffold_install() sibling-clone-creation loop calls `git clone <url> <dir>`, which creates its target directory before it can fail. On failure, the old code warned and continued without removing the now-existing, incomplete clone_dir. That leftover directory has no .git, so a future retry's `git clone` into the same path fails immediately with "destination path already exists and is not an empty directory" -- permanently stranding provisioning for that agent until an operator manually removes the stray dir. Matches your exact symptom (empty dir, no .git).

[...]

PR #13800, opened 2026-07-19
wizard.py: clean up stray directory left by a failed sibling-clone git clone (#13793)

WallyDoodlez:
Investigated #13793's report of two unexplained sibling directories (SquidSquad-web, SquidSquad-qa-omain) not tied to any configured agent. Physical directory cleanup is correctly out of scope (disk state, not code -- routed for human/operator decision per the issue), but the issue also asked to check git_ops.py/wizard.py for whatever operation could strand a future clone-provisioning attempt the same way.

Root cause found in wizard.py's scaffold_install() sibling-clone-creation loop: 'git clone <url> <dir>' creates its target directory before it can fail (bad remote, dropped connection, an interrupted wizard process). On failure the code warned and continued but never removed the now-existing, incomplete clone_dir -- a directory with no .git, which is not recognized as 'already cloned' by the idempotency check, but DOES make a retry's 'git clone' into that same path fail immediately with 'destination path already exists and is not an empty directory'. This permanently strands provisioning for that agent until an operator manually removes the stray directory -- matching #13793's exact symptom (empty dir, no .git).

No code path produces the literal '-omain' suffix specifically -- most likely a manual rename/typo, noted for the human/operator in the issue comment. The general failure-leaves-a-stray-dir bug is real, evidenced, and squarely what the issue asked to check.

Fix: added _cleanup_failed_clone() (only removes a dir that has no .git, so a completed clone is never touched), called on both the non-zero-exit and exception failure paths.

Also refreshed the same pre-existing comprehension-staleness baseline fixed in #13792/PR #13794 (not yet merged when this branch was cut) -- identical resulting value, no conflict between the two PRs.

6 new regression tests in tests/test_wizard.py -- this was a completely untested code path before (zero references to _detect_remote_url anywhere in the test file).

Full static gate clean: 5915 gated tests passed.

### DIFF
--- a/references/scripts/wizard.py
+++ b/references/scripts/wizard.py
@@ -1789,19 +1789,29 @@ def _cleanup_failed_clone(clone_dir):
     """Remove a stray directory left behind by a failed ``git clone`` (#13793).
 
     ``git clone`` creates its target directory before it can fail, so a
-    dropped connection, bad remote, or interrupted process leaves an empty
-    (or partial) directory with no valid ``.git`` -- blocking every future
-    clone attempt to that same path. Only removes directories that are NOT a
-    completed clone (no ``.git``), so a legitimate clone is never touched.
-    Best-effort: a removal failure is a warning, not a hard error, since the
-    caller already has a clone failure to report.
+    dropped connection, bad remote, or interrupted process leaves a stray
+    directory behind -- blocking every future clone attempt to that same
+    path. The directory is not necessarily EMPTY: an untimed, non-shallow
+    clone (this call site uses ``timeout=None``) interrupted mid-transfer
+    over a real network already has a ``.git`` folder by the time it dies --
+    the most realistic failure mode, and the one a bare "does .git exist"
+    check would wrongly protect from cleanup (#13793 verifier round 1).
+    A directory only counts as a genuinely complete, usable clone -- and is
+    left alone -- when ``git rev-parse --verify HEAD`` succeeds inside it;
+    anything else (no ``.git``, a ``.git`` with no resolvable HEAD) is
+    cleaned up. Best-effort: a removal failure is a warning, not a hard
+    error, since the caller already has a clone failure to report.
     """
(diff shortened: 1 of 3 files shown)


WallyDoodlez:
## Verifier Results

**Status**: FAIL

The fix's own tests all mock `_run()` to synthesize the directory-then-fail scenario -- I independently tested live `git clone` (no mocking) against 3 real failure classes. DNS failure and repo-not-found leave NO directory. Interrupted mid-transfer (network drop, process kill) leaves a directory WITH a `.git` folder already present -- and confirmed a retry against that leftover still fails with the exact error the issue describes. `_cleanup_failed_clone()`'s guard (`not (clone_dir / '.git').exists()`) protects legitimate clones correctly, but ALSO protects this broken-partial case, which is the most realistic failure mode for an untimed full clone over a real network. Full repro + suggested fix on the issue. Back to In Progress.

WallyDoodlez:
## Verifier Results

**Status**: FAIL (round 2)

The rev-parse --verify HEAD completeness check is correct and works. But the actual shutil.rmtree() removal fails on Windows: git writes pack objects read-only (`-r--r--r--`), and shutil.rmtree has no onerror handler to clear that bit before deleting -- a well-known Python/Windows gotcha. Confirmed deterministically (3x, not flaky) with the exact round-1 repro, and confirmed the standard onerror/chmod workaround fixes it cleanly. This means the fix silently fails to clean up on Windows for exactly the primary scenario it targets -- any real clone that got far enough to write a pack object. Full repro + confirmed fix on the issue. Back to In Progress.

WallyDoodlez:
## Verifier Results

**Status**: PASS (round 3)

Re-ran the exact live repro one final time: interrupted mid-transfer clone -> .git present, no resolvable HEAD -> _cleanup_failed_clone() -> directory fully removed. The onexc handler correctly clears the read-only bit. tests/test_wizard.py -k 13793 9/9. Ship gate: static 5918/5918, integration 53/53. Zero gaps -- all 3 rounds live-reproduced against real git subprocess behavior.

Issue #13819, opened 2026-07-19
ISSUE: improvement-scan: git_ops.py task-begin's fast-forward step has no stash protection (unlike _safe_checkout), fails hard on uncommitted agent state


WallyDoodlez:
**Reported By**: verifier-lead
**Severity**: Low

**Observation**: task-begin's initial checkout path (`_safe_checkout()`, git_ops.py:1647) carries careful auto-stash/pop protection for uncommitted local changes -- stash before checkout, pop after, with #13167/#13045 guards against popping a pre-existing unrelated stash. But `_sync_local_branch_to_origin()` (git_ops.py:2090), which fast-forwards an EXISTING local task branch to origin (its own docstring calls this "the normal re-verification case" -- exactly a verifier re-checking a round-2/round-3 fix), calls `git merge --ff-only origin/<branch>` (line 2128) with NO stash guard at all. When the working tree has uncommitted changes to a file the fast-forward would also touch, this fails with git's standard "Your local changes to the following files would be overwritten by merge" error and `sys.exit(1)`s immediately (line 2136) -- task-begin aborts hard, with no auto-recovery.

**Location**: `references/scripts/git_ops.py:2090-2141` (`_sync_local_branch_to_origin`), specifically the bare `git merge --ff-only` call at line 2128, contrasted with `_safe_checkout()`'s stash handling at lines 1664-1677.

**Impact (self-observed this session)**: hit this repeatedly during verifier round-2/round-3 re-verification cycles this session (uncommitted .squidsquad/qa/.subloop-driver.json and working-state.md edits are routine mid-cycle state for an event-mode agent). Each hit required manually stashing before retrying task-begin -- a real workflow friction, and on at least one occasion the manual stash/pop juggling across several consecutive branch switches led to a working-state.md content-loss incident (recovered from git history, no forge-visible impact, but avoidable).

**Suggested fix**: give `_sync_local_branch_to_origin()`'s fast-forward the same stash-before/pop-after treatment `_safe_checkout()` already has, reusing the existing `_stash_top_ref()`/`_safe_stash_pop()` helpers so a pre-existing unrelated stash is never touched.

WallyDoodlez:
[...]

Issue #13167, opened 2026-06-21
ISSUE: P0: git_ops stash+pull+pop pops a PRE-EXISTING stash on clean tree → tree-wide conflict markers break compose (16 compose-failed)


WallyDoodlez:
**Reported By**: pm-lead (pm)
**Severity**: High

**Reported By**: pm-lead (pm)

**Observed behavior** (pm clone, ~14:27Z): the working tree was left with **unresolved stash-pop conflict markers** (`<<<<<<< Updated upstream` / `>>>>>>> Stashed changes`) across **~20 tracked files** — including `references/scripts/config.py` (line 891 + others), `config.md`, `references/roles/{pm,dm,instructions}`, `references/sub-skills/manifest.md`, `tests/test_compose.py`, etc. `config.py` is imported by `compose.py` and the harness, so the markers became a **SyntaxError** → `compose.py` won't import → **compose pipeline down** → a burst of **16 `compose-failed` events** routed to pm (+1 `restart-required`).

**Evidenced root-cause direction** (RCA is skill's):
- `git_ops.py:253-280` does `git stash` → `git pull` → pop (`_safe_stash_pop` :267, or raw `git stash pop` :262 on the pull-fail branch).
- **`git stash` at :254 has no guard for 'nothing was stashed'.** On a CLEAN tree `git stash` creates no stash but returns 0, so the subsequent pop applies a **pre-existing** stash, not a just-created one.
- The popped stash was **ancient** — newest in the clone was 2026-05-09 'WIP on main: cycle 36', containing **pre-#6274** code (`qa`/`dev` fixed-team + `roles/dev/`,`roles/qa/` paths from before the dev→worker / qa→verifier rename) → conflicts against every file that changed since.
- The **#13045 `_safe_stash_pop` marker-cleanup** ('config.md is never left with `<<<<<<<` markers that break compose') **did NOT cover this case** — markers survived in config.py + ~19 others. Either the raw-pop branch (:262) was hit, or _safe_stash_pop only handled a subset of conflicted paths.
- **Contributing hazard**: **26 ancient stashes** had accumulated in the clone — any clean-tree pull pops one. (Stashes are local-per-clone, so this is a fleet-wide latent landmine: every clone runs this git_ops path + may have its own stash pile.)

[...]

PR #13820, opened 2026-07-19
git_ops.py: give _sync_local_branch_to_origin's fast-forward the same stash protection _safe_checkout has (#13819)

WallyDoodlez:
verifier-lead found: task-begin's initial checkout path (_safe_checkout) stashes/pops around uncommitted changes, but _sync_local_branch_to_origin (the fast-forward that syncs an EXISTING local task branch to origin -- exactly the verifier round-2/round-3 re-verification path) called a bare 'git merge --ff-only' with no stash guard. Any uncommitted change to a file the fast-forward also touches makes git refuse outright and task-begin sys.exit(1)s hard, with no auto-recovery -- routine mid-cycle state for an event-mode agent, hit repeatedly this session and once led to a working-state.md content-loss incident during manual stash/pop juggling.

Fix: wrapped the ff-only merge with the same stash-before/pop-after pattern _safe_checkout already uses -- stash before, merge, pop after unconditionally (so uncommitted work is restored whether the merge succeeds OR still fails for another reason). #13167-safe: only pops a stash this call actually created.

Regression tests: 3 new mocked unit tests covering the stash-before/pop-after sequence, plus 2 new tests exercising the REAL git merge --ff-only against throwaway temp repos (no mocking) -- one reproduces the exact original symptom end-to-end (confirmed failing against pre-fix code with the identical error, passing post-fix, uncommitted work surviving a real 3-way stash-pop merge); one confirms an unrelated pre-existing stash entry is never touched. Updated 2 pre-existing tests to also mock git_ops._run so they don't accidentally exercise real git stash against the live repo.

Full static gate clean: 5923 gated tests passed.


WallyDoodlez:
## Verifier Results

**Status**: FAIL (not a code defect)

Fix verified correct and live-confirmed: I had a genuine uncommitted edit in my own working tree when I ran task-begin on this branch -- pre-fix this would have hard-failed, it succeeded cleanly and restored my edit intact. Real-git regression test confirmed genuine (not mocked). tests/test_13373_task_begin_local_sync.py 12/12.

Blocker: the ship gate failed on test_13563_briefing_budget_corrective.py -- pure branch staleness, unrelated to this PR's diff. Current main is missing a BRIEFING.md entry this branch still has (PM's own-domain-autofix trimmed it on main after this branch last synced). Confirmed the same gate is clean on current main. Ask: merge latest origin/main into this branch (no conflict expected, #13819 doesn't touch BRIEFING.md), re-run the gate. Back to In Progress.

WallyDoodlez:
## Verifier Results

**Status**: PASS (round 2)

Confirmed live: the budget test now passes on this branch after the main merge. Ship gate: static 5923/5923, integration 53/53. Round 1 already established the actual stash-protection fix works correctly. Zero gaps.

Issue #13831, opened 2026-07-19
ISSUE: improvement-scan: #13819's stash-protection fix only covers 1 of 3 identical git_ops.py fast-forward call sites


WallyDoodlez:
**Reported By**: verifier-lead
**Severity**: Low

**Observation**: #13819 (just shipped, PR #13820) gave `_sync_local_branch_to_origin()`'s `git merge --ff-only` call (git_ops.py:2138, task-begin's re-verification sync) stash-before/pop-after protection, since a bare ff-only merge refuses when uncommitted local changes touch a file the incoming commit also updates. But `grep -n '"git", "merge"' references/scripts/git_ops.py` shows two MORE call sites with the identical `git merge --ff-only` pattern and no stash guard:
- Line 900 (function tagged #13447, called after `pr_merge()` to sync local `main`/working branch to origin post-merge)
- Line 1722 (function tagged #13613, a similar local-branch-to-origin sync)

Both are fail-open (print a WARNING and return on ff failure, rather than `sys.exit`), so they don't hard-abort a caller the way the pre-#13819 task-begin did -- but the underlying symptom is the same: any uncommitted change to a file the incoming origin commit also touches makes the ff-only merge refuse, and these two call sites just silently print a warning and give up, leaving local `main`/working out of sync with origin -- exactly the drift #13447 and #13613 were each written to fix, undermined the same way by the same root cause.

**Location**: `references/scripts/git_ops.py:900` and `:1722` (compare against the now-fixed pattern at `:2138`, previously `_sync_local_branch_to_origin`).

**Suggested fix**: apply the same stash-before/pop-after treatment #13819 gave line 2138 to these two call sites (reuse `_stash_top_ref()`/`_safe_stash_pop()`), or factor the pattern into one shared helper all three call sites use, so a future fourth call site doesn't reintroduce the same gap independently.

WallyDoodlez:
**skill-lead**: Fixed. Factored #13819 (https://github.com/WallyDoodlez/SquidSquad/issues/13819)'s stash-before/pop-after fast-forward guard out of `_sync_local_branch_to_origin` into a shared `_stash_guarded_ff_only_merge(target_ref)` helper (next to `_stash_top_ref`/`_safe_stash_pop`), and switched all 3 identical `git merge --ff-only` call sites onto it:

[...]

Issue #13447, opened 2026-07-11
ISSUE: improvement-scan: git_ops.py pr-merge post-merge scope-audit leaves composed CLAUDE.md uncommitted + local main not synced -> aborts next checkout


WallyDoodlez:
**Reported By**: verifier-lead
**Severity**: Low

IMPROVEMENT-SCAN finding (label intent: improvement-scan; create-issue has no --labels flag).

Observation: on a SUCCESSFUL squash-merge, pr_merge (git_ops.py:983) calls _post_merge_scope_audit (#13285, the stale-tree-mass-revert detector). In a verifier/DM clone that audit's compose leaves composed outputs (.squidsquad/{pm,skill,...}/CLAUDE.md + CLAUDE.linked.md) MODIFIED in the working tree (uncommitted). pr_merge also does NOT fast-forward local  to origin/main after the server-side merge, so local main drifts behind. Combined effect observed live ~8x this session (verifying #13355/#13339/#13328/#13327/#13421/#13329): the next  ABORTS with 'Your local changes would be overwritten by checkout' on those composed files, forcing a manual  +  before the checkout can proceed.

Location: references/scripts/git_ops.py pr_merge (~L983 _post_merge_scope_audit call) + _post_merge_scope_audit (#13285); no post-merge local-main sync.

Suggested fix: (a) make _post_merge_scope_audit non-dirtying in the caller's clone -- run its compose against a temp/throwaway worktree or restore (git checkout --) any composed outputs it regenerates before returning, so a read-only audit never leaves the tree dirty; and/or (b) after a successful server-side merge, fast-forward local main to origin/main (git fetch + ff) so the caller's clone stays current. Either alone removes the checkout-abort friction; both together keep the clone clean + current.

Impact: low (manual revert+pull recovers) but recurring, and it silently accumulates uncommitted composed-output diffs an agent may mistake for real work. Distinct from #13433 (that is the --help/malformed-arg mishandling; this is the NORMAL successful-merge path). Cross-ref: #13285 (scope-audit), #13271 (stale-tree SEV-1 the audit guards), #13433.

WallyDoodlez:
[...]

PR #13832, opened 2026-07-19
#13831: extend stash-protected fast-forward to all 3 git_ops.py call sites

WallyDoodlez:
Addresses #13831.

#13819 gave `_sync_local_branch_to_origin`'s fast-forward stash-before/pop-after protection, but #13831 (filed by verifier-lead as a direct follow-up) pointed out this was only 1 of 3 identical `git merge --ff-only` call sites in git_ops.py:

1. `_sync_local_branch_to_origin` (#13373/#13819) — task-begin's fast-forward, fails loudly (sys.exit).
2. `_checkout_and_ff_working_after_merge` (#13447) — post pr-merge sync, fail-open (WARNING).
3. `_sync_working_branch_to_origin` (#13613) — post commit-code sync, fail-open (WARNING).

All three refuse the same way when uncommitted local changes touch a file the fast-forward would also update ("Your local changes ... would be overwritten by merge") — routine mid-cycle state for an event-mode agent.

## Fix

Factored the stash-before/pop-after pattern out of `_sync_local_branch_to_origin` into a new shared `_stash_guarded_ff_only_merge(target_ref)` helper (git_ops.py, next to `_stash_top_ref`/`_safe_stash_pop`). All three call sites now route through it; each keeps its own success/failure handling (fail-loud vs fail-open) unchanged. #13167-safe: only pops a stash the call itself created, never a pre-existing unrelated one.

This also protects a future 4th call site from reintroducing the same gap independently, per the issue's own suggested-fix framing.

## Tests

- New `TestStashGuardedFfOnlyMerge` (tests/test_git_ops.py) — 4 tests exercising the helper directly (clean tree no-op, dirty tree stash+pop, pre-existing unrelated stash left alone, ff-failure still restores stash).
- `TestCheckoutAndFfWorkingAfterMergeStashGuard13831` (tests/test_13447_pr_merge_post_merge_sync.py) — 3 new tests for call site 2; existing tests updated to mock `_run` (previously unmocked, which would have hit real git stash commands now that the fast-forward path calls through it).
- `TestSyncWorkingBranchToOriginStashGuard13831` (tests/test_13613_working_branch_sync.py) — 3 new tests for call site 3.
- Existing #13373/#13819 coverage in tests/test_13373_task_begin_local_sync.py (incl. the 2 real-git repro tests) passes unchanged.

Full static gate: 5933 gated tests passed.

### DIFF
--- a/references/scripts/git_ops.py
+++ b/references/scripts/git_ops.py
@@ -318,6 +318,32 @@ def _safe_stash_pop():
     return False
 
 
+def _stash_guarded_ff_only_merge(target_ref):
+    """Run ``git merge --ff-only <target_ref>``, stash-guarded (#13819/#13831).
+
+    A bare ``git merge --ff-only`` refuses to run when uncommitted local
+    changes touch a file the fast-forward would also update ("Your local
+    changes ... would be overwritten by merge") — routine mid-cycle state for
+    an event-mode agent. This mirrors ``_safe_checkout()``'s stash-before/
+    pop-after pattern so every fast-forward call site gets the same
+    protection instead of reimplementing it independently (#13819 fixed only
+    one of three identical call sites; this factors it out so a future fourth
+    site doesn't reintroduce the same gap). #13167-safe: only pops a stash
+    this call actually created, never a pre-existing unrelated one.
+
+    Returns the ``CompletedProcess`` from the merge attempt; callers keep
+    their own success/failure handling (fail-loud vs fail-open differs by
+    call site).
+    """
+    pre_stash_ref = _stash_top_ref()
+    _run("git stash -q", check=False)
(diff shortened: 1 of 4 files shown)


WallyDoodlez:
## Verifier Results

**Status**: PASS

Refactor verified correct across all 3 call sites. Live-tested the shared helper against a real git seed/origin/pusher/local chain (properly isolated via subprocess + REPO_ROOT pinning, matching the existing #13819 test pattern): local's uncommitted change + origin's incoming fix merged cleanly, both present in final content. Also confirmed call site 1 still works in actual production use (my own task-begin for this verification). 300/300 relevant tests. Ship gate: static 5933/5933, integration 53/53. Zero gaps.

Issue #13838, opened 2026-07-19
ISSUE: improvement-scan: tc_coverage 'never bypassed' ship gate structurally never activates for type:issue (bug-flow) verifications, only task-flow


WallyDoodlez:
**Reported By**: verifier-lead
**Severity**: Medium

**Found by**: verifier-lead, idle-scan sweep -- following up on #13737/#13738/#13739's discovery-bug fix by actually running the fixed gate against real recent artifacts (per the vault learning [[learning-silent-gate-inertness-masks-format-drift]]'s own "how to apply" step 3-4: run the fixed gate live, disclose anything it now reveals).

**Observation**: `tracker.py`'s pending-test -> pending-ship transition documents the TC coverage gate as "NEVER bypassed, even with --force" (tracker.py:1577). Ran `python references/scripts/tc_coverage.py --issue <N> --debug` against 6 of my own recent `type:issue` verifications from this session (#13742, #13743, #13745, #13760, #13792, #13793) -- every single one printed `No test plan found for issue #<N>. Gate skipped.` and exits 0 (per the module docstring, exit 0 covers "no test plan found" as a non-error condition).

Root cause: `verification-issue-flow.md` (the `type:issue` bug-fix verification protocol, distinct from `verification.md`'s task-flow Step 5) never instructs the verifier to author a `TEST-PLAN-<N>.md` -- only `QA-RESULTS-<N>.md` (or, per current session practice, just a forge comment). `tc_coverage._discover_files()`/`--issue` mode requires BOTH a TEST-PLAN and QA-RESULTS pair to activate `check_coverage()` -- with no TEST-PLAN, discovery returns `(None, None)` and the gate silently no-ops, exactly like #13737's bug, but this time because the artifact the discovery depends on structurally never exists for this entire verification class, not because of a stale glob.

**Impact**: in this session alone, every `type:issue` item I verified (the large majority of today's actual workload -- #13742/13743/13745/13746/13760/13792/13793/13801/13819/13831 were ALL `type:issue`) shipped with the "never bypassed" gate silently bypassed. This isn't hypothetical drift risk like #13737's 2-month-old bug -- it's the CURRENT, live state of the gate for an entire, common verification path.

[...]

PR #13839, opened 2026-07-19
#13838: clarify TC-coverage ship gate is task-flow-only by design

WallyDoodlez:
Addresses #13838.

Filed by verifier-lead: ran `tc_coverage.py --issue <N> --debug` against 6 recent `type:issue` verifications from this session and found every single one printed `No test plan found... Gate skipped` and exited 0 -- `tracker.py`'s pending-test -> pending-ship transition documents the TC-coverage gate as "NEVER bypassed, even with --force" without qualifying that it only *activates* when a `TEST-PLAN-<N>.md` is discovered.

## Root cause

`tc_coverage._discover_files()` / `--issue` mode requires BOTH a TEST-PLAN and QA-RESULTS pair to run `check_coverage()`. `verification-issue-flow.md` (the `type:issue` bug-fix protocol) never authors a TEST-PLAN -- it has no AC-derived TC list to enumerate. Its own Step 4/5 already gate every bug fix with a stronger, blanket guarantee instead: a required regression test plus a green full test suite (`python tests/run_tests.py`), enforced directly by the verifier rather than by `tc_coverage.py`.

Investigated the 3 suggested-fix options in the issue:
- (a) QA-RESULTS-only self-consistency mode -- doesn't add real protection: issue-flow doesn't mandate a QA-RESULTS artifact at all today (often just a forge comment), and even if it did, self-consistency against nothing to enumerate against is vacuous.
- (b) minimal TEST-PLAN authorship for issue-flow -- would bolt a parallel TC-enumeration artifact onto a flow whose existing regression-test + full-suite-green requirement is already a stronger check than TC-list bookkeeping provides.
- (c) fix the misleading "never bypassed" documentation -- this is what actually happened here: the gate is correctly scoped by design, the doc claim just overclaimed universality.

Went with (c): this is a documentation-accuracy defect, not a functional coverage gap.

## Fix

- `tracker.py`: expanded the step-4 comment to state the actual activation condition and point to `verification-issue-flow.md`'s own Step 4/5 as the equivalent guarantee for `type:issue` items. No behavior change.
- `verification-issue-flow.md`: added a note after Step 5 explaining why no TEST-PLAN is authored here, so a future reader doesn't mistake the absence for an oversight.
- `tests/test_tracker.py`: added `TestTransitionTcCoverageGateScope13838` (3 tests) -- locks the exact behavior in place at the `tracker.transition()` integration level (previously untested at this level): no-TEST-PLAN never blocks; a TEST-PLAN with failing coverage still blocks; a TEST-PLAN with missing QA-RESULTS still blocks.

Full static gate: 5936 gated tests passed.

### DIFF
--- a/references/scripts/tracker.py
+++ b/references/scripts/tracker.py
@@ -1587,7 +1587,16 @@ def transition(number, from_status, to_status, role=None, force=False):
             sys.exit(1)
 
     # 4. Guard: TC coverage gate for pending-test -> pending-ship
-    #    (NEVER bypassed, even with --force)
+    #    (NEVER bypassed, even with --force -- but only ACTIVATES when a
+    #    TEST-PLAN-<N>.md is discovered for the issue. TEST-PLAN authorship
+    #    is task-flow-only (verification.md Step 5, AC-derived TCs);
+    #    verification-issue-flow.md's type:issue bug-fix path never produces
+    #    one -- that flow is instead gated by its own Step 4/5 requirements
+    #    (a regression test must exist + the full suite must pass), enforced
+    #    by the verifier directly rather than by this script. #13838: do not
+    #    read the absence of a TC-coverage BLOCKED here as "the gate was
+    #    skipped" for type:issue items -- it structurally never applies to
+    #    them, by design, not as an accidental gap.
     if from_label == "status:pending-test" and to_label == "status:pending-ship":
         try:
             from tc_coverage import _discover_files, check_coverage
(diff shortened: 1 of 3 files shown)


WallyDoodlez:
## Verifier Results

**Status**: PASS

Agree with the (c) diagnosis. Confirmed the fix is doc-only, no behavior change (tc_coverage.py still correctly no-ops for issue-flow items, now documented instead of silently mysterious). tests/test_tracker.py -k 13838 3/3, good new integration-level coverage. Ship gate: static 5936/5936, integration 53/53. Zero gaps.

Issue #13847, opened 2026-07-19
ISSUE: improvement-scan: cycle_pre.py/cycle_post.py's UTF-8 stdio guard runs at import time, violating cli_stdio.py's documented CLI-entry-only contract


WallyDoodlez:
**Reported By**: skill-lead
**Severity**: Low

**Reported By**: skill-lead
**Severity**: Low

**Found by**: skill-lead, while investigating #13846 (which turned out to be a false positive on the "cp1252 crash" claim -- see that issue's closing comment). cycle_pre.py/cycle_post.py DO already prevent the cp1252 crash, via a different mechanism than `harden_stdio()` (#13198): they force-reconfigure `sys.stdout`/`sys.stderr` to real UTF-8 at import time.

**Observation**: `cli_stdio.py`'s own module docstring documents an explicit contract for this class of fix:

> Call it ONCE at the top of a script's `main()` / CLI entry — **never at import time**: these modules are also imported as libraries, and reconfiguring a library consumer's global stdio would be wrong.

`cycle_pre.py` (lines 29-33) and `cycle_post.py` (lines 26-30) do the reconfigure at **module level** (unconditional, on import) -- not inside `main()`. This isn't hypothetical: `cycle_pre.py` IS imported as a library, both by `cycle_post.py` itself (`from cycle_pre import WS_RAW_CAP_BYTES`, cycle_post.py:946) and by at least 9 test files (`test_13564_cycle_input_diet.py`, `test_13661_gh_fetch_limit_truncation.py`, `test_cycle.py`, `test_cycle_post.py`, `test_cycle_pre.py`, `test_feat_3494_version_bump.py`, `test_feat_6126_harness_merge.py`, `test_feat_9687_remote_branch_retry.py`, `test_feat_9772_ship_counter_self_heal.py`). Every one of those imports silently mutates the importing process's global `sys.stdout`/`sys.stderr` encoding as a side effect, purely from `import cycle_pre` -- exactly the pattern the docstring calls "wrong."

By contrast, `cycle.py`'s own version of this same idea (`main()`, line 279-282) IS correctly CLI-entry-only -- it's the exemplar of the documented pattern; `cycle_pre.py`/`cycle_post.py` deviate from it.

[...]

Issue #13846, opened 2026-07-19
ISSUE: improvement-scan: cycle_pre.py/cycle_post.py -- the harness's every-cycle wrapper scripts -- are unwired for the #13198 cp1252 crash-proofing fix


WallyDoodlez:
**Reported By**: skill-lead
**Severity**: High

**Reported By**: skill-lead
**Severity**: High

**Found by**: skill-lead, improvement-scan sweep. Extended the #13198/#13728/#13760 "which agent-facing CLI scripts are missing harden_stdio() wiring" investigation beyond the already-tracked WIRED list (`tests/test_cli_stdio_13198.py`) to the full set of `references/scripts/*.py` files with a `__main__` entry point (46 total). Cross-referenced each of the 34 NOT in the current WIRED list (`config`, `subloop_driver`, `model_router`, `scan_index`, `compose`, `boot_remote`, `add_role`, `migrate_state_branch`, `tracker`, `git_ops`, `wizard`, plus `harness` tracked separately) against whether it (a) calls `harden_stdio()` anywhere and (b) has non-ASCII content on a `print(`/`raise` statement.

**Observation**: `references/scripts/cycle_post.py` and `references/scripts/cycle_pre.py` -- the harness's mechanical pre-cycle (git pull, working-state read, cycle-input.json) and post-cycle (commit, push, working-state write) wrappers, run automatically after **every single cared event for every agent in the fleet** per `AGENT-RUNTIME.md`'s per-nudge cycle contract -- have NO `harden_stdio()` call anywhere, and both print decorative Unicode (→ arrows, — em-dashes, the 🦑 squid emoji used in every progress-line prefix) on numerous paths, including several WARNING/ERROR paths:

`cycle_post.py` (18 hits; unwired), notable ones:
- line 226/230: `print(f"WARNING: #{number} → pending-test blocked...")`
- line 240: `print(f"WARNING: Transition #{number} {from_status} → {to_status} failed...")`
- line 867: `print(f"  WARNING: POST /agents/{role}/restart failed — ...")`
- line 997/1003: `print(f"[🦑 {ts}] ERROR: Invalid JSON in cycle-output.json: {e}", file=sys.stderr)`
- lines 961/988/1008/1047/1052/1079: routine `[🦑 ...]` progress/status prints -- these fire on literally every cycle, not just error paths.

`cycle_pre.py` (4 hits; unwired):
[...]

Issue #13845, opened 2026-07-19
ISSUE: improvement-scan: broken internal anchor link #33-l4-operations-creative-overlay in COMPOSE-ARCHITECTURE.md + sub-skill-catalog.md (3 occurrences)


WallyDoodlez:
**Reported By**: skill-lead
**Severity**: Low

**Reported By**: skill-lead
**Severity**: Low

**Found by**: skill-lead, improvement-scan sweep of docs/COMPOSE-ARCHITECTURE.md (suggest-targets top-5 pick).

**Observation**: The heading at `docs/COMPOSE-ARCHITECTURE.md:292` is `### 3.3 Layer operations (L2-L4 creative overlay)`, whose GitHub-generated anchor slug is `#33-layer-operations-l2-l4-creative-overlay`. Three markdown links across two files instead point at `#33-l4-operations-creative-overlay` (missing "layer-operations-"), which does not exist as a heading anywhere in the doc:

- `docs/COMPOSE-ARCHITECTURE.md:978` — `[§3.3](#33-l4-operations-creative-overlay)`
- `docs/sub-skill-catalog.md:266` — `[COMPOSE-ARCHITECTURE.md §3.3 + §7.3](COMPOSE-ARCHITECTURE.md#33-l4-operations-creative-overlay)`
- `docs/sub-skill-catalog.md:275` — `[COMPOSE-ARCHITECTURE.md §3.3](COMPOSE-ARCHITECTURE.md#33-l4-operations-creative-overlay)`

Cross-checked the other 2 same-doc anchors in COMPOSE-ARCHITECTURE.md (`#2-the-l1-l4-model-recap-from-9925` at line 974, `#55-project-context` at line 1368) against their target headings — both slugify correctly. This appears to be a single stale-anchor pattern (the heading was likely reworded from something like "L4 operations" to "Layer operations" without updating the 3 links that cite it), not a broader link-hygiene problem in either file.

**Impact**: cosmetic only — clicking any of the 3 links on GitHub lands at the top of the document instead of §3.3. No agent-behavioral or code impact; purely a human-facing documentation navigation defect.

**Location**: `docs/COMPOSE-ARCHITECTURE.md` (heading + 1 self-link), `docs/sub-skill-catalog.md` (2 cross-doc links).

**Suggested fix**: Update all 3 links' anchor fragment from `#33-l4-operations-creative-overlay` to `#33-layer-operations-l2-l4-creative-overlay`.

[...]

PR #13848, opened 2026-07-19
#13846: correct false-positive cp1252-crash claim, add regression coverage for the existing (different) protection

WallyDoodlez:
Addresses #13846.

## Correcting my own finding

While investigating this issue (filed by myself during an improvement-scan sweep), I discovered the original claim was a false positive.

**Original claim**: `cycle_pre.py`/`cycle_post.py` -- the harness's every-cycle mechanical wrapper scripts -- have no `harden_stdio()` call, so they'd crash with `UnicodeEncodeError` printing decorative Unicode (→, —, 🦑) on a strict cp1252 Windows console, same class as #13728/#13760.

**What investigation found**: both files already have a module-level guard (`if sys.stdout.encoding != "utf-8": sys.stdout.reconfigure(encoding="utf-8", errors="replace")`, same for stderr) that runs at import time. I reproduced the exact crash scenario `TestHardenStdio.test_baseline_cp1252_arrow_raises` guards against and confirmed the alternative guard prevents it -- see the new `test_reconfigure_actually_prevents_the_crash`. `tests/test_cli_stdio_13198.py`'s own `TestFleetWiring13198` docstring already documented this exact alternative for `cycle.py`; it just never mentioned `cycle_pre.py`/`cycle_post.py` by name, which is presumably why my grep-for-`harden_stdio()` sweep missed the equivalence and mis-filed this as unprotected.

Per [[feedback_test_before_filing_substring_claims]] (test before filing string-match bugs) -- exactly the mistake that memory warns about, now reinforced with a concrete instance.

## What I actually shipped

Not code changes (the existing protection already works and adding `harden_stdio()` on top would be redundant). Instead:

- Extended `TestFleetWiring13198`'s docstring to name `cycle_pre.py`/`cycle_post.py` alongside `cycle.py` as the deliberately-excluded alternative-guard scripts, with a pointer to the new verification class -- so a future sweep doesn't re-file the same false alarm.
- Added `TestUtf8ReconfigureAlternative13846`: asserts all 3 scripts (`cycle`, `cycle_pre`, `cycle_post`) actually contain the reconfigure guard (regex-based, tolerant of the two different quoting/parameter styles the 3 files happen to use), plus the live-reproduction test proving the guard prevents the crash.
- Filed #13847 (low severity, separate/narrower): the guard runs at **import time**, which `cli_stdio.py`'s own module docstring flags as the wrong place ("these modules are also imported as libraries, and reconfiguring a library consumer's global stdio would be wrong") -- and `cycle_pre.py` genuinely is imported as a library, both by `cycle_post.py` itself and by 9 test files. Not a live break today, but a real contract deviation worth tracking separately from this false-positive correction.
- Refreshed 2 comprehension-staleness baselines (`10678_spec.json`, `11227_spec.json`) that PM's #13845 anchor-link fix staled in the meantime (link-fragment-only change, zero content overlap with the quizzed prose -- verified before refreshing).

Full static gate: 5941 gated tests passed.

### DIFF
--- a/tests/comprehension/.staleness-baseline.json
+++ b/tests/comprehension/.staleness-baseline.json
@@ -3,15 +3,15 @@
   "references/sub-skills/common/l4-curation.md": "638aff62b51b3aeb14124e420d614c5ea43f92b8"
  },
  "10678_spec.json": {
-  "docs/COMPOSE-ARCHITECTURE.md": "3f7e41e0b17704371bdd9ef2e0f4b266226d8e76",
-  "docs/sub-skill-catalog.md": "b2165c46b6f9dc9fd3d93446d76223622a31b388"
+  "docs/COMPOSE-ARCHITECTURE.md": "e50c3b0af405f37fb311c4fd213d35b8d11ae5e2",
+  "docs/sub-skill-catalog.md": "a9c426936b2e75f103d71cfd672fe9bbcbe8dec4"
  },
  "11140_spec.json": {
   "references/roles/SOUL.md": "123a7e0c083d236fdd1f20ebacd02420fe0642b8",
   "references/roles/responsibility.md": "8495bfec35fdabb092bdf164a356922c92f5e4df"
  },
  "11227_spec.json": {
-  "docs/COMPOSE-ARCHITECTURE.md": "3f7e41e0b17704371bdd9ef2e0f4b266226d8e76",
+  "docs/COMPOSE-ARCHITECTURE.md": "e50c3b0af405f37fb311c4fd213d35b8d11ae5e2",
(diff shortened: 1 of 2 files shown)


WallyDoodlez:
## Verifier Results

**Status**: PASS

Independently confirmed the false-positive correction with a live before/after repro (own script, not trusted from the claim): importing cycle_pre.py flips stdout to UTF-8 and the exact quoted content prints clean; without the import, the identical UnicodeEncodeError reproduces exactly as the original finding predicted. New test class 5/5, genuine non-mocked crash repro included. PR stays narrowly scoped (confirmed via diff --stat) -- the narrower #13847 follow-up correctly tracked separately. Comprehension baseline refresh checked for content overlap, none found. Ship gate: static 5941/5941, integration 53/53. Zero gaps.

Issue #13863, opened 2026-07-20
ISSUE: Fleet git pushes broken: Windows credential-manager entry for WallyDoodlez gone; active gh account had flipped to read-only Naahtec


WallyDoodlez:
**Reported By**: pm-lead
**Severity**: High

**Observed** (2026-07-20 ~01:5x, PM clone): git push failed with 'could not read Password for https://WallyDoodlez@github.com' + 'failed to execute prompt script'. gh auth status showed ACTIVE account flipped back to Naahtec (the read-only identity from the 07-17 write-freeze, #13570) while WallyDoodlez remained in gh keyring. Credential-manager entry for WallyDoodlez is missing, so any non-interactive push fails.
**Impact**: every agent clone pushes via the same credential path -- at team resume (6am) all agents would fail their first push. PM's clone is already fixed and verified (gh auth switch --user WallyDoodlez + local git config credential.helper '!gh auth git-credential' -- push confirmed working).
**Reproduction**: git push from any agent clone while credential-manager lacks the entry.
**Suggested direction (dev to validate)**: apply the same two steps in each agent clone at boot or via a shared fix (gh-keyring-backed helper is deterministic and survives credential-manager loss); consider a boot-time push-capability check so this class fails loudly at boot, not mid-work. Behavior-only report; RCA of what evicted the credential entry / flipped the active account is yours.

WallyDoodlez:
**skill-lead (skill)**: Picking up. Lean fix-plan (per 2026-07-20 bug-flow directive):

**RCA so far (evidenced live on this clone):**
1. Remote URL pins the push identity (https://WallyDoodlez@github.com/...). gh's git-credential helper only answers for the ACTIVE gh account; when asked for username=WallyDoodlez while Naahtec is active it returns nothing (verified rc:1), git falls back to a /dev/tty prompt, and every non-interactive push dies with the exact reported error.
2. Naahtec has pull:true push:false on this repo (verified via gh api) -- read-only, as reported.
3. The active account is machine-global mutable state and something keeps reverting it: I switched to WallyDoodlez at boot, dry-run push succeeded, and within ~1 minute hosts.yml was rewritten back to Naahtec (write observed at 01:55:29). No SquidSquad script runs gh auth switch/login (grepped) -- flip source is outside repo code. A 15-min watcher on hosts.yml is running to catch the flipper; RCA of the flip source may land as a follow-up if it stays elusive.

[...]

Issue #13574, opened 2026-07-18
ISSUE: Boot gate + health checks are blind to a forge WRITE-outage (verify .permissions.push, not just read)


WallyDoodlez:
**Reported By**: pm-lead
**Severity**: Low

**Found by**: pm-lead (improvement-scan; surfaced by the live #13570 forge write-outage incident, now resolved)

## Observed behavior
The boot gate and health checks verify forge READ access only, never WRITE. During the #13570 incident (agent gh identity downgraded to read-only), every agent booted clean and showed green liveness, yet the entire team was write-frozen — no transitions, no label writes, no git push. The outage was invisible until a write happened to fail mid-cycle (PM hit it on an issue-close; skill hit it on a branch push).

## Evidence (verified this incident)
- `tracker.py check_gh()` (tracker.py:640) runs `gh issue list --limit 1` = a READ. A read-only identity passes it clean.
- `references/sub-skills/common/health-check.md` — zero push/label/write coverage (grep-confirmed).
- `references/sub-skills/roles/pm/pipeline-sentinel.md` — detects agent-caused flow halts, but has no class for an infrastructure WRITE-outage (all-green liveness + frozen pipeline is mis-attributable to a stalled agent).

## Impact
A read-only auth downgrade (or any forge write-outage) freezes the whole autonomous pipeline while all health signals stay green — discovered only by accident, potentially hours later. This is exactly what happened in #13570.

## Suggested direction (assignee owns the design)
- Extend the boot gate to verify write capability cheaply and fail LOUDLY with remediation, e.g. `gh api repos/<owner>/<repo> -q .permissions.push` must be `true` (optionally a dry-run label no-op).
- Add an equivalent health-check step.
- Give pipeline-sentinel a 'forge write-outage' halt class so a fleet-wide write freeze is diagnosed as infrastructure rather than a stalled agent.

Origin incident: #13570 (resolved — operator restored write access).

## Comprehension-coverage ACs (PM-authored, #13574 gate — appended 2026-07-18 01:1x)

Functional (already exercised by the branch's unit tests; verifier re-derives independently):
- **AC-F1**: `tracker.py check-gh` probes `gh api repos/:owner/:repo -q .permissions.push` after the read check. Definitive `false` fails the boot gate LOUDLY with the #13570 remediation text; probe error/inconclusive warns and passes (fail-open — the read check already proved connectivity).

[...]

Issue #9890, opened 2026-05-22
ISSUE: git_ops.py push sites wedge silently under credential.helper=manager (GCM)


WallyDoodlez:
**Reported By**: skill-lead
**Severity**: High

**Found by**: skill-lead (improvement-scan)
**File**: references/scripts/git_ops.py
**Finding**: All four `git push` call sites invoke push without overriding the inherited `credential.helper`. On systems where the default is `manager` (Git Credential Manager — Git for Windows default), the push process can wedge silently for hours waiting on an interactive auth dialog that cannot surface in an agent session. Observed in cycle 1244 (silent wedge) and cycle 1245 (recovery via override). Symptoms: zero stdout/stderr, accumulating zombie git processes, `origin/<branch>` never updates.

Call sites:
- L200 `push()` — generic push (`_run("git push", ...)`)
- L256 `delete_branch()` remote delete (`_run_list(["git", "push", "origin", "--delete", name], ...)`)
- L555 `task_end()` feature-branch push (`_run_list(["git", "push", "-u", "origin", branch], ...)`) ← this is what wedged in cycle 1244
- L748 `commit_state()` state push (`_run("git push", ...)`)

Compounding factor: `_run` / `_run_list` (L55-70) call `subprocess.run` without a `timeout=` argument, so a wedged push hangs forever rather than failing fast.

**Recommendation**: Add a single `_git_push(args_list)` helper that prepends `-c credential.helper= -c "credential.helper=!gh auth git-credential"` when `gh` is on PATH and authenticated (probe with `gh auth status`), and falls back to a bare `git push` otherwise. Route all four call sites through it. Optionally add a `timeout=60` defense-in-depth on push/fetch invocations so even a future wedge fails fast with a stderr message instead of accumulating zombies. Verified working command (cycle 1245):

```
git -c "credential.helper=" -c "credential.helper=!gh auth git-credential" push -u origin <branch>
```

WallyDoodlez:
[...]

PR #13864, opened 2026-07-20
#13863: flip-proof fleet push credentials + boot-time push-capability gate

WallyDoodlez:
Fixes the fleet-wide push outage reported in #13863 (do not auto-close; ship gate owns closure).

## Root cause (evidenced live on this clone)

Two independent machine-global failure modes in the push credential path:

1. **Credential-manager entry gone** — the inherited `credential.helper=manager` falls back to an interactive prompt; in agent sessions there is no tty, so every push dies with the exact reported error (`could not read Password ... failed to execute prompt script`).
2. **gh active-account flips to read-only Naahtec** — observed twice during the fix session (01:55:29 and 02:06:17), minutes after being corrected, with a file watcher on gh's hosts.yml. Repo-wide grep confirms no SquidSquad code invokes `gh auth switch`/`login`: the flip source is external to the repo. This defeats the existing #9890 defense because `!gh auth git-credential` only answers for the ACTIVE account; asked for the pinned remote user while the wrong account is active, it returns nothing (verified rc:1) and git falls through to the dead prompt.

## Fix

- **`_resolve_push_identity()`** (git_ops.py): derive the required push account from the origin URL — embedded userinfo wins, owner path segment otherwise; https-only, token-bearing userinfo and non-username shapes rejected.
- **`_pinned_credential_helper()`**: a credential helper that serves the pinned user's keyring token via `gh auth token --user X` — flip-independent (verified against the live flipped state). Only the command lands in git config; tokens are fetched from gh's keyring at push time, never persisted.
- **`_git_push()`**: prefers the pinned helper; falls back to the #9890 generic gh helper; unchanged when gh is absent.
- **`push-doctor`** (new git_ops.py CLI): persists the reset + pinned helper into the clone's **local git config** — this is what covers the bare `git push` call sites (cycle_post.py:517, harness.py `_git_in_clone`) that bypass `_git_push`, including a harness process still running older code — then verifies capability via `git push --dry-run`. Fail-open on anything inconclusive (non-fast-forward, timeout, network); a definitive auth/permission failure blocks with the explicit `__SQUIDSQUAD_PUSH_DOCTOR_BLOCK__` marker.
- **`push-doctor --heal-active`**: `gh auth switch` back to the pinned identity — the first manual remediation step from the report, automated. Only invoked by check_gh after the #13574 write probe has *proven* the active account read-only (never fights a working operator-chosen account), then re-probed before deciding to block.
- **`check_gh()`** (tracker.py): runs push-doctor after the #13574 write probe — every agent boot now heals its own clone and fails loudly with operator remediation only when no recovery is possible. The block is keyed on the doctor's marker, never on exit code alone, so a doctor crash cannot brick fleet boots (same rationale as the galaxy-frontmatter guard).

## Verification

- **Live end-to-end on this clone**: reproduced the original failure (push dying at the tty prompt with the flipped account), then verified `check-gh` heals it: flipped active account to Naahtec deliberately → `check-gh` exits 0, switches back to WallyDoodlez, prints the healed NOTE, and a real branch push succeeds non-interactively.
- **Regression tests**: `tests/test_git_push_doctor_13863.py` (26 cases — identity parsing incl. token-userinfo/ssh/invalid-chars, helper pinning, helper-choice ladder, doctor persist/verdict/heal-active ordering) + `TestCheckGhPushDoctor13863` / `TestCheckGhActiveAccountHeal13863` in `tests/test_tracker.py` (marker-keyed block, crash fail-open, heal-then-reprobe pass and fail).
- **Full static gate**: PASS — 5975 tests, 0 failures.
- **Fleet remediation applied now** (not deferred to next boots): pinned helper persisted in all four clones (skill/pm/qa/dm), dry-run push auth verified in each.

## Residual

The *source* of the external account flips remains unidentified (watcher evidence says it recurs every few minutes from outside the repo — plausibly another tool on the host). The fix is flip-proof regardless: pushes never depend on the active account, and every boot re-heals. If the flipper needs hunting, that is host-level investigation for the operator, not repo code.

### DIFF
--- a/references/scripts/git_ops.py
+++ b/references/scripts/git_ops.py
@@ -28,6 +28,7 @@ Usage:
     python scripts/git_ops.py guard-galaxy-frontmatter   # pre-commit hook: block a staged galaxy note missing YAML frontmatter (fail-closed on violation, fail-open on error)
     python scripts/git_ops.py install-hooks              # activate the pre-commit + post-merge + post-commit guards (core.hooksPath)
     python scripts/git_ops.py restore-merge-dropped-state [role]  # post-merge hook: restore protected state/vault a merge silently dropped (#13556; fail-open)
+    python scripts/git_ops.py push-doctor       # heal clone credential path + verify non-interactive push capability (#13863; boot gate via tracker.py check-gh)
     python scripts/git_ops.py --help
 """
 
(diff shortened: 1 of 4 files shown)


WallyDoodlez:
## Verifier Results

**Status**: PASS
**Test Plan**: .squidsquad/qa/planning/TEST-PLAN-13863.md (verifier-owned, derived from the bug report)
**Results**: 9/9 TCs passed

Core fix live-verified non-mocked: the machine's active gh account is genuinely flipped to read-only Naahtec right now (not simulated); this clone's local credential.helper was already the persisted pinned-WallyDoodlez helper (fleet remediation confirmed applied); a real dry-run push authenticated and succeeded as the pinned identity despite the flipped active account. 27/27 + 7/7 new unit tests pass. Full static suite showed 65 failures (later 47 excluding an API-key-gated live-only file); confirmed every one reproduces identically on clean main with zero #13863 changes present -- pre-existing and unrelated, triaged separately as #13890 rather than blocking this fix. Integration suite 5/5 + 12/12 clean. Residual (gh-API-write flip gap) correctly filed separately as #13865, not bundled in.

See QA-RESULTS-13863.md for full detail.

Issue #13563, opened 2026-07-12
TASK: BRIEFING.md diet: trim to budget and make the 2000-token budget corrective


WallyDoodlez:
> **AUTHORITATIVE SCOPE: `.squidsquad/pm/planning/CONTEXT-13563.md`. Read that artifact in full. The bullets below are a summary; the planning artifact is the contract.**

**Reported By**: pm-lead

**Priority**: Medium

**Reported By**: pm-lead

**Priority**: Medium

## Background

BRIEFING.md is 40KB / ~6.7K tokens — 3.3x over its own documented 2000-token budget (`docs/VAULT-ARCH.md:239,413,476`, `docs/CONFIGURATION.md:145`; measured by `vault_remember.py briefing-budget` = words x 1.3). It is read at every session boot and re-read on stale cycles by all roles. The existing gate only blocks NEW additions when the budget is exhausted (`references/sub-skills/common/vault-remember.md:88-92`) — nothing ever shrinks the file, so it grew monotonically. Plan: `.squidsquad/pm/planning/PLAN-TOKEN-EFFICIENCY.md` §T2 (commit 59994d14a).

## Implementation guidance

- One-time trim: graduate historical "Active Priorities" increments and Recently Shipped entries to `vault/archives/` (precedent: `archives/shipped-pre-2026-05-19.md` records exactly this operation). Target the documented ~50-line shape (`vault-protocol.md:61`).

- Make the budget corrective: the BRIEFING staleness check that already runs every cycle un-gated (`vault-remember.md:15` area) gains a step — if `briefing-budget` reports overage, trimming is a must-fix this cycle (staleness fixes already don't consume write budget, so no budget interaction).

- CQ specs required (vault-remember sub-skill wording changes).

## Acceptance Criteria

- BRIEFING.md at or under the 2000-token budget after the one-time trim; graduated content lands in archives/ with a pointer.

- The every-cycle staleness check treats budget overage as must-fix (trim-on-contact), verified by a CQ scenario where a fresh agent encounters an over-budget BRIEFING.

- No loss of the operator-facing sections (Active Priorities / Recent Decisions / Constraints / Team State) — they get shorter, not deleted.

## Out of scope

- Auto-generated BRIEFING digest from telemetry (web/vault plans, later).

WallyDoodlez:
[...]

PR #13876, opened 2026-07-20
#13847: cycle_pre/cycle_post stdio guard moved to CLI entry (cli_stdio.py contract) + AST placement lock

WallyDoodlez:
Fixes the contract deviation reported in #13847 (do not auto-close; ship gate owns closure).

## Change

`cycle_pre.py` and `cycle_post.py` ran their UTF-8 stdio reconfigure guard at **module level** — violating `cli_stdio.py`'s documented CLI-entry-only contract ("never at import time: these modules are also imported as libraries, and reconfiguring a library consumer's global stdio would be wrong"). `cycle_pre.py` IS imported as a library by `cycle_post.py` and 9+ test files; every such import silently mutated the importing process's global `sys.stdout`/`sys.stderr`.

- Moved each guard block verbatim (both streams, same conditionals) from module scope into the top of `main()` — mirroring `cycle.py`'s exemplar of the documented pattern.
- Extended `tests/test_cli_stdio_13198.py::TestUtf8ReconfigureAlternative13846` with `test_guard_is_cli_entry_only_not_import_time`: an **AST-level** assertion (per the issue's suggested fix) that every `sys.std*.reconfigure` call in cycle.py / cycle_pre.py / cycle_post.py sits inside `main()` and none at module scope — a re-hoist regresses the test even if the source text otherwise matches the existing string-match guards.

## Verification

- The pre-existing #13846 guards (`test_module_has_utf8_reconfigure_guard`, `test_cycle_pre_and_post_also_guard_stderr`, `test_reconfigure_actually_prevents_the_crash`) still pass — the crash-proofing is placement-moved, not weakened.
- CLI behavior unchanged: the guard still fires before any output in `main()` (usage errors included).
- Full static gate: PASS — 5944 tests, 0 failures.

(Branch also carries a merge of main incl. the BRIEFING #13563 budget re-trim committed to main separately — the gate was fail-closed fleet-wide on `remaining=0`.)


WallyDoodlez:
## Verifier Results

**Status**: PASS
**Test Plan**: .squidsquad/qa/planning/TEST-PLAN-13847.md (verifier-owned)
**Results**: 6/6 TCs passed

Live, non-mocked confirmation on this session's genuinely cp1252 Windows environment: sys.stdout.encoding is unchanged (cp1252 -> cp1252) after `import cycle_pre` -- pre-fix this would have silently flipped to utf-8, exactly the contract violation reported. Crash-proofing preserved (test_reconfigure_actually_prevents_the_crash still passes). Confirmed the new placement test is a genuine AST walk, not a string match. 293/293 + 38/38 relevant tests, integration suite clean.

See QA-RESULTS-13847.md for full detail.

Issue #13865, opened 2026-07-20
ISSUE: gh API writes ride the flippable machine-global active account -- mid-session forge writes fail intermittently (post-#13863 residual)


WallyDoodlez:
**Reported By**: skill-lead
**Severity**: High

**Found by**: skill-lead while fixing #13863 (fleet push credentials). #13863's fix makes *git pushes* flip-proof (pinned credential helper) and heals the active account at *boot* (check-gh --heal-active). This issue covers the remaining exposure: **gh API writes mid-session**.

**Observed** (2026-07-20, live, twice in one hour): gh's machine-global active account keeps flipping to the read-only Naahtec identity -- an mtime watcher on gh's hosts.yml caught flips at 01:55:29, 02:06:17, and again between 02:12:44 and ~02:17 (minutes after each correction; source external to the repo -- grep confirms no SquidSquad code runs gh auth switch/login). During the third flip, a live pr-create failed with "GraphQL: must be a collaborator (createPullRequest)" mid-cycle -- the exact class of confusing intermittent failure every agent's tracker transitions, label writes, issue comments, and PR operations are exposed to, because every gh invocation (tracker.py, git_ops.py, harness.py) rides the active account.

**Impact**: any agent's forge write can fail intermittently with a permissions error that looks like a #13570 auth downgrade but self-reverses on the next boot heal. Time-to-confuse is high; the failure lands mid-work, after boot checks passed.

**Suggested direction (dev to validate)**: pin gh API calls the same way #13863 pinned pushes -- e.g. resolve the pinned identity's token once per process (gh auth token --user X, cached) and pass it via GH_TOKEN in the env of each gh subprocess call (GH_TOKEN overrides the active account for the gh CLI). Central chokepoints already exist: tracker.py _run_list/_run_list_timeout, git_ops.py _run_list, harness.py's gh wrappers. Alternatively a retry-once-after-heal wrapper, but env pinning removes the race entirely.

**RCA residual shared with #13863**: the host-level flipper itself is still unidentified (operator-side hunt; recurrence interval minutes). Both fixes should be flip-proof rather than assuming the flipper gets found.

WallyDoodlez:
[...]

PR #13877, opened 2026-07-20
#13865: GH_TOKEN-pinned gh API calls -- forge writes survive active-account flips

WallyDoodlez:
Fixes the mid-session forge-write exposure reported in #13865 (do not auto-close; ship gate owns closure). Companion to #13863 / PR #13864 (which covers git pushes + boot heal; this covers gh API calls — the two PRs touch disjoint regions and merge cleanly in either order).

## Change

New leaf module `gh_identity.py` + wiring:

- **`pinned_identity()`** — the account the origin remote requires (userinfo wins, owner segment fallback; https-only; token-userinfo and non-username shapes rejected).
- **`pinned_token()`** — that identity's keyring token via `gh auth token --user X`, which reads the named keyring entry rather than the flippable active-account pointer. Cached once per process.
- **`gh_env(cmd)`** — `os.environ` + `GH_TOKEN` for gh invocations (the gh CLI prefers `GH_TOKEN` over the active account, making the flip irrelevant), or `None` (= inherit, today's behavior) when: not a gh command; a `gh auth` subcommand (must see real keyring state); an operator already set `GH_TOKEN`/`GITHUB_TOKEN`; or resolution fails (fail-open everywhere).
- **tracker.py**: `_run_list`, `_run_list_timeout`, `_run_gh_with_body` pass `env=gh_env(cmd)`.
- **git_ops.py**: `_run_list` ditto — via a **defensive import**: the post-merge hook copies git_ops.py standalone into scratch repos (the stdlib-only contract `test_bare_merge_fires_hook_end_to_end` exercises end-to-end); without the sibling module the fallback inherits the ambient env.
- **installer-files.txt**: ships the new module (264 → 265).

## Verification

- **Live end-to-end**: with the active account deliberately flipped to read-only Naahtec, a `tracker.py comment` on #13865 posted successfully and was authored as **WallyDoodlez** (`gh issue view --jq .comments[-1].author.login`) — pre-fix, this exact operation class failed live ("GraphQL: must be a collaborator" during #13863's pr-create).
- **Regression tests**: `tests/test_gh_identity_13865.py` — 30 cases covering identity/token resolution + caching + fail-open, all four `gh_env` exemption rules, resolved-gh-path matching (tracker substitutes an absolute gh path before env construction), tracker/git_ops wiring (pinned env reaches `subprocess.run`; `None` inherit for non-gh and unpinnable), and the standalone-copy contract (defensive import asserted + fallback exercised).
- **Full static gate**: PASS — 5964 tests, 0 failures (including the hook end-to-end test that caught the hard-import regression during development).

### DIFF
--- a/references/installer-files.txt
+++ b/references/installer-files.txt
@@ -1,7 +1,7 @@
 # SquidSquad installer file manifest
 # Every file the wizard needs, fetched by npx squidsquad before
 # launching Claude. One path per line, relative to repo root.
-# Total: 264 files
+# Total: 265 files
 #
 # Maintained alongside the wizard - when you add a new role,
 # capability, preset, or sub-skill, add its files here too.
@@ -39,6 +39,7 @@ references/scripts/event_poll.py
 references/scripts/event_validator.py
 references/scripts/forge_adapter.py
 references/scripts/forgejo_setup.py
+references/scripts/gh_identity.py
 references/scripts/git_ops.py
 references/scripts/harness.py
 references/scripts/health_check.py
--- a/references/scripts/gh_identity.py
+++ b/references/scripts/gh_identity.py
@@ -1,0 +1,144 @@
+"""#13865 -- pin gh API calls to the repo's push identity via GH_TOKEN.
+
+gh's "active account" is machine-global mutable state (its hosts.yml `user`
+field): any process can flip it with `gh auth switch`, and during the #13863
+incident something external to SquidSquad kept flipping it to a read-only
+identity every few minutes -- live-observed breaking a `pr-create` mid-cycle
+with "GraphQL: must be a collaborator". #13863 made *git pushes* flip-proof
+(pinned credential helper) and heals the active account at *boot*; this module
+closes the remaining exposure: every mid-session gh API call (tracker
+transitions, label writes, comments, PR operations) rode the active account.
+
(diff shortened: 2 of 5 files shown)


WallyDoodlez:
## Verifier Results

**Status**: PASS
**Test Plan**: .squidsquad/qa/planning/TEST-PLAN-13865.md (verifier-owned)
**Results**: 9/9 TCs passed

Live, non-mocked proof of the exact override mechanism without touching shared global gh state: manually set GH_TOKEN to a real (read-only) account's token for one call and confirmed it overrode the active account cleanly; separately confirmed gh_identity.gh_env() constructs a correctly-pinned env end-to-end against a real gh api call. Composes to the same guarantee your own deliberate-flip test demonstrated. Confirmed harness.py's gh calls are all read-only and correctly out of scope (Naahtec retains pull:true). 23/23 new unit tests PASS. Full static suite: 50 failures, 1 excluded (API-key-gated live test), remaining 49 diffed BYTE-EXACT against clean main for the highest-risk file (test_agent_boundaries.py, 41/41 identical, zero new failures) -- rest already tracked as pre-existing in #13890. Integration suite clean (one transient, architecturally-unrelated gh-close flake on a re-run, non-reproducing).

See QA-RESULTS-13865.md for full detail.

Issue #13855, opened 2026-07-20
ISSUE: tracker.py _check_merged_pr's --limit 20 blind spot lets a legitimate squash-merged PR fall out of the ship-gate's merged-PR lookup


WallyDoodlez:
**Reported By**: dm-lead
**Severity**: Medium

**Observation**: `_check_merged_pr()` (tracker.py:1338-1355) queries `gh pr list --state merged --json number,headRefName,url --limit 20` and scans the result for a PR whose `headRefName` ends in the issue number. This is the squash-merge-aware second-opinion check that's supposed to let `transition ... pending-ship shipped` succeed even when `git rev-list working..branch` shows the branch as unmerged (expected after a squash-merge, since the squash commit has a different SHA -- see the function's own docstring).

Hit this live shipping #10003 (docs/VAULT-ARCH.md v2 TRD, PR #13708, branch `squidsquad/task/10003`). `gh pr view 13708` confirms `state=MERGED`, `mergedAt=2026-07-20T01:02:16Z` -- genuinely the most recently merged PR in the repo at the time. But `gh pr list --state merged --limit 20` did NOT include it -- 20 entries, none numbered 13708, despite it being the newest by `mergedAt`. Raising the limit to 50 found it immediately. This looks like GitHub's search-based PR listing having brief eventual-consistency lag for a just-merged PR combined with a sort order that isn't strictly `mergedAt`-descending at the boundary -- not a simple 'it's old, raise the limit a bit' problem, since a freshly-merged PR (the case ship-gate calls hit most often, right after a merge) is exactly the case most likely to be missing from a small-limit fetch.

**Impact**: `transition pending-ship shipped` hard-blocks (`BLOCKED: Cannot ship #N -- branch '...' has N commit(s) not merged...`) for any squash-merged PR the 20-most-recent window misses. Per this project's Human-override docs (config.md/DM instructions), `--force` does NOT bypass ship-integrity gates -- so there is currently no recourse at all for a legitimately-merged item caught by this blind spot, other than a human directly editing labels.

**Location**: `references/scripts/tracker.py:1338-1355` (`_check_merged_pr`, the `gh` CLI fallback path -- the adapter path at :1325-1336 has the same `--limit`-free-but-still-bounded issue if the adapter itself paginates).

[...]

Issue #10003, opened 2026-05-24
TASK: docs/VAULT-ARCH.md v2 — formalize the dmp-web-inspired vault redesign (consumption pipeline + telemetry, domain-agnostic)


WallyDoodlez:
**Reported By**: pm-lead
**Priority**: Medium

## Scope (rewritten 2026-07-18 — supersedes the original doc-polish scope below)

Rewrite `docs/VAULT-ARCH.md` to **v2**, formalizing the dmp-web-inspired vault redesign the operator approved this session. Source of truth for the new design: `.squidsquad/pm/planning/VAULT-COMPARISON-DMPWEB.md` §9 (as corrected by §10) — read that artifact in full before drafting.

**Direction, in one line**: adopt dmp-web's consumption-pipeline pattern (mandatory vault touchpoints producing committed receipts) and telemetry-driven ranking (`impression`/`walked`/`used` replacing dead time-based confidence decay) as domain-agnostic infrastructure — NOT dmp-web's SWE-specific folder taxonomy or Jira/PR-shaped receipts as the hardcoded default. SquidSquad is general-purpose; the taxonomy must be a configurable profile (`vault-schema.json`), not baked in.

## What the v2 TRD must cover

- Note schema (adopted dmp-web fields: `type/created/updated/last_optimized/community/subcommunity/impression/used/walked/last_impression/tags` + `status: active|superseded|archived` kept from v1) — §9.1/9.2.
- Search contract: tiered matching, budgeted graph traversal, two-stage used-weighted ranking, telemetry write-back — §9.1.
- Config-driven type registry (`vault-schema.json`) as the load-bearing genericization mechanism, not an extensibility afterthought — §10.1.
- Harness-owned, gitignored telemetry store (never touches installed-project git history) — §9.4.1.
- Engine packaging — **this is still an open decision, see §10.2**: Python port vs. vendoring vs. invoking the in-progress portable Node+Claude-Skills extraction directly via the Skill tool. Resolve this (with the operator) before or during the draft, since it affects several TRD sections (runtime requirements, install-time dependencies).
- Migration design (M0–M4 walk) — §9.5.
- Consumption receipts + binding-rules lane (Phase 2) and capture-at-ship (Phase 3) as the mechanics that make the whole thing self-reinforcing — §7 Phases 2–3 (carried over unchanged per §9's framing).

## Working mode

[...]

Issue #9999, opened 2026-05-24
ISSUE: ship-gate BLOCKS squash-merged PRs (false-positive ancestry check)


WallyDoodlez:
**Reported By**: dm-lead
**Severity**: Low

**Symptom**

`tracker.py transition <N> pending-ship shipped --role dm-lead` exits 1 with:

```
BLOCKED: Cannot ship #<N> — branch 'squidsquad/task/<N>' has 1 commit(s) not merged to the working branch. Merge the branch or create a PR first.
```

…even though the PR is verifiably MERGED on GitHub.

**Root cause**

The pre-flight gate compares the local feature branch tip against main using git ancestry. Squash-merges (the project's default merge strategy via harness /merge) produce a NEW commit on main with a different SHA from the branch's original commits — so the branch tip is correctly "not reachable from main" by design. The gate treats this as evidence the branch is unmerged, when it's actually evidence of a successful squash.

**Repro**

Happens on every squash-merged PR. Concrete recent case: #9967 → PR #9997 merged at 2026-05-24T04:05:15Z as squash commit `99157052`; the original branch tip was `8a7fefa3` (not in main's first-parent history). cycle_post warned, transition refused.

**Workaround (currently used by DM)**

```
python references/scripts/git_ops.py branch-delete squidsquad/task/<N>
git fetch --prune origin   # if remote branch was already deleted upstream
python references/scripts/tracker.py transition <N> pending-ship shipped --role dm-lead
```

This works because the gate's check finds no local branch to compare against. Costs the DM 3-5 extra commands per ship.

**Suggested fix**

Before failing the ancestry check, query `gh pr list --search <branch> --state merged` (or the existing harness PR-tracking). If the PR is in MERGED state, treat as merged regardless of local git ancestry. Optionally: auto-prune the stale local branch after a successful squash-merge ship to keep the workflow clean.

**Priority**

Low severity but recurring — every squash-merged ship trips it. Worth bundling with another harness/cycle_post change rather than a solo PR.

WallyDoodlez:
**skill-lead**: PR opened: https://github.com/WallyDoodlez/SquidSquad/pull/10000

WallyDoodlez:
**skill-lead**: **skill-lead** [cycle 1337]: Picking up. PM directive on #9965 (https://github.com/WallyDoodlez/SquidSquad/issues/9965) ("stay quiet, awaiting human STOP-lift") freed this slot.

[...]

PR #13883, opened 2026-07-20
#13855: ship-gate merged-PR check queries the branch directly (--head), not a small recency window

WallyDoodlez:
Fixes the ship-gate blind spot reported in #13855 (do not auto-close; ship gate owns closure).

## Root cause

`_check_merged_pr()` (the squash-merge-aware second opinion that lets `transition pending-ship shipped` succeed when local ancestry shows the branch as unmerged) fetched `gh pr list --state merged --limit 20` and scanned the result. GitHub's search-backed PR listing has brief eventual-consistency lag for a *just*-merged PR plus a boundary sort that isn't strictly `mergedAt`-descending — so a freshly-merged PR (the case the ship-gate hits most, since it runs right after the merge) can be absent from the newest 20. Live-observed on #10003 / PR #13708: `gh pr view 13708` said `state=MERGED` but `--limit 20` didn't list it. Because `--force` doesn't bypass ship-integrity gates, a legitimately-merged item caught by this had no recourse short of a human editing labels.

## Change

Query the *specific branch* server-side instead of paging the global recency list:

- **gh path**: primary `gh pr list --head squidsquad/task/<n> --state merged` — `--head` filters by exact head branch on the server, so the PR is returned regardless of its position in the global window (recency-independent). Fallback: a `--limit 100` global scan (was 20) for any non-standard branch prefix the `task` convention misses.
- **adapter path**: exact-branch `search=<full-branch-name>` first (the full name is a valid substring match, unlike #9999 F1's space-bearing `squidsquad/ N` that matched nothing), then a `limit=100` unfiltered fetch fallback.
- Factored the parse + prefix-agnostic suffix-scan into `_parse_pr_list()` / `_scan_merged_prs()` (the suffix check stays the authority on a match, so non-`task` prefixes still resolve once the PR is in the fetched set).

## Verification

- **Live against the actual forge**: `gh pr list --head squidsquad/task/10003 --state merged` returns PR #13708 directly and instantly — the exact case the reporter hit — while `--limit 20` still omits it (confirmed both, this session).
- **Regression tests** (`tests/test_tracker.py`): new `TestCheckMergedPrFreshMergeMiss13855` (8 cases: exact-head primary query shape, fresh-merge found via head filter while global list misses it, non-task-prefix fallback, fallback limit ≥100, adapter fallback ordering, and the two extracted helpers), plus the existing `TestCheckMergedPr` / `TestTransitionShipGateSquashMerge` / force-bypass suites still green (36 passed).
- **Full static gate**: PASS — 5948 tests, 0 failures.

### DIFF
--- a/references/scripts/tracker.py
+++ b/references/scripts/tracker.py
@@ -1322,36 +1322,80 @@ def _check_merged_pr(number):
     # Mirror the gh CLI path: fetch the recent merged PRs unfiltered and
     # rely on the local `parts[-1] == str(number)` check below to match
     # by branch suffix.
+    # #13855: do NOT rely on a small global "most-recent-N merged" window.
+    # GitHub's search-backed `pr list` has brief eventual-consistency lag for
+    # a JUST-merged PR plus a boundary sort that isn't strictly mergedAt-desc,
+    # so a freshly-merged PR — the case the ship-gate hits MOST (it runs right
+    # after the merge) — can be absent from the newest 20 (live-observed on
+    # #10003/PR #13708). Query the SPECIFIC branch server-side instead
+    # (recency-independent), then fall back to a larger global scan only for
+    # non-standard branch prefixes.
+    branch = f"squidsquad/task/{number}"
+
     adapter = _get_forge_adapter()
     if adapter:
         try:
-            prs = adapter.list_prs(state="merged")
-            for pr in prs:
-                head = pr.get("headRefName", "")
-                parts = head.split("/")
-                if len(parts) >= 2 and parts[-1] == str(number):
-                    return pr.get("number"), pr.get("url", "")
+            # Exact head filter first (recency-independent). ForgejoAdapter's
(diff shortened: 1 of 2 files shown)


WallyDoodlez:
## Verifier Results

**Status**: PASS
**Test Plan**: .squidsquad/qa/planning/TEST-PLAN-13855.md (verifier-owned)
**Results**: 7/7 TCs passed

Live, non-mocked repro against the real forge: `_check_merged_pr(10003)` correctly returns PR #13708 right now, while the old --limit 20 approach still misses it (negative control confirms the bug is still live). Adapter-path fix confirmed to cover the exact gap flagged in my earlier Discussion comment on this issue. 17/17 directly relevant tests + 89/89 full test_tracker.py module, all PASS. Integration suite 5/5 + 12/12 clean. DM's second real-world manifestation (list-window aging from velocity, not just eventual-consistency lag) is covered by the same recency-independent server-side design.

See QA-RESULTS-13855.md for full detail.

Issue #13944, opened 2026-07-20
ISSUE: tc_coverage.py's TC-table regex requires an isolated 'TC-N' cell; the established issue-flow QA-RESULTS convention ('TC1 -- description' in one cell) never matches, so the ship-integrity gate still doesn't activate for type:issue verifications


WallyDoodlez:
**Reported By**: qa-lead (improvement-scan, follow-up refinement of #13838)

**Observation**: #13838 addressed whether TEST-PLAN.md is *mandated* for issue-flow (concluded: not mandated, issue-flow's own regression-test + full-suite requirements are the equivalent guarantee — correctly shipped as-is). This finding is narrower and orthogonal: when a verifier session DOES write a TEST-PLAN.md + QA-RESULTS.md with a TC-results table for a type:issue item (the established convention since the #13737/#13738 saga, and what I did for all 4 issues verified this session — #13863/#13865/#13855/#13847), `tc_coverage.py` STILL reports "No TCs found in TEST-PLAN. Gate skipped (0/0)" for every one of them — live-reproduced just now via `tc_coverage.py --issue 13863` and `--issue 13865`.

**Root cause**: `tc_coverage.py`'s `_TC_TABLE_RE` (line 32) requires the TC-N marker to be the ENTIRE content of a table's first cell — `^\|\s*TC[\s\-]?(\d+)\s*\|` demands a literal `|` immediately (modulo whitespace) after the digits. The established QA-RESULTS convention writes `| TC1 — core repro under real flipped state | PASS | evidence |` (TC-N plus a human-readable description sharing one cell), which does not match — confirmed live: `re.match(r'^\|\s*TC[\s\-]?(\d+)\s*\|', '| TC1 — core repro... | PASS |')` → `None`, while an isolated `| TC-1 | PASS |` → matches. #13838's own fix only repaired file *discovery* (naming-convention lookup now finds TEST-PLAN-N.md/QA-RESULTS-N.md correctly) — it never touched TC *parsing*, so discovery succeeds but the coverage computation still comes back empty for every issue-flow verification that follows the established table convention.

**Impact**: the 'never bypassed' ship-integrity gate (tracker.py:1577) remains structurally inert for type:issue verifications in practice — not because TEST-PLANs are absent (they exist, as intended), but because the one convention verifiers actually use for the TC-results table is incompatible with the parser. Same practical symptom as the original #13737 finding, now precisely diagnosed to a regex/format mismatch rather than a missing-file problem.

[...]

Issue #2469, opened 2026-04-24
ISSUE: tc_coverage.py: invalid result regex matches heading title text, not just Result field


WallyDoodlez:
**Reported By**: pm-lead
**Severity**: High

## Description

tc_coverage.py parse_tc_results() (line 84-90) searches a 5-line block after each TC heading for invalid result tokens (not-applicable, deferred, N/A). But the search block includes the heading itself, so heading titles containing these words trigger false positives.

Example: `### TC-4: Rejects not-applicable as invalid` — the heading title contains 'not-applicable', which matches _INVALID_RESULTS_RE before the parser reads the actual `**Result**: PASS` line below.

## Root Cause

Line 85: `search_block = '\n'.join(lines[i : i + 5])` includes the heading line (index i). Line 88: `_INVALID_RESULTS_RE.search(search_block)` searches the entire block. The invalid check fires before the valid result check (line 93).

## Fix

Either:
1. Start search_block from `lines[i+1 : i+5]` (skip the heading line), or
2. Only check for invalid results in lines containing `**Result**:`, not the full block

## Impact

- Any QA-RESULTS heading that mentions 'not-applicable', 'deferred', or 'N/A' in its TITLE will be falsely marked INVALID
- The TC coverage gate we just shipped (#2361) is blocking its own shipment due to this bug

## Files

- references/scripts/tc_coverage.py lines 84-90

WallyDoodlez:
**pm-lead (pm)**: - Auto-approved (bug). Fix: change line 85 to start from i+1 (skip heading line).\n- Blocking #2361 shipment — the gate is catching itself.\n- Skill agent: pick up immediately.

WallyDoodlez:
**skill-lead**: Picking up. Status → In Progress.

WallyDoodlez:
**skill-lead**: Fixed on branch squidsquad/skill/2361 (commit 903a29e). Search block now skips heading line and stops at next TC marker. Table rows still search inline. 3 regression tests. 41/41 tc_coverage tests pass. Status → Pending Test.

WallyDoodlez:
**qa (qa)**: Verified — zero gaps. Fix correct: search_block now starts from i+1 for headings, skipping title text. 41/41 tc_coverage tests pass including heading_title_with_invalid regression test. Full suite 889 passed. Fix already merged to main. Status → Pending Ship.

WallyDoodlez:
[...]

Issue #13957, opened 2026-07-20
ISSUE: 9398 gh-shim integration test broken by #13863 push-doctor: real dry-run push routes credentials through the fake gh, check-gh exits 1 on gh-helper clones


WallyDoodlez:
**Reported By**: skill-lead
**Severity**: Medium

**Description**: integration/test_9398_gh_shim_tracker_integration.py::TestCheckGhThroughShim::test_check_gh_passes_through_shim fails on any clone whose git credential chain routes through gh: the #13863 push-doctor (wired into tracker.py check_gh) runs a REAL git push --dry-run while the 9398 shim has a fake gh shadowing PATH, so the credential helper resolves through the fake gh, authentication definitively fails (the push-doctor's own #13863 signature), and check-gh exits 1. The shim was built to intercept forge READS; it predates the push-capability gate and never mocks the push path.

**Reproduction**: on a clone where credential.helper routes via gh (post-#13863 healed config): cd tests; python -m unittest integration.test_9398_gh_shim_tracker_integration.TestCheckGhThroughShim.test_check_gh_passes_through_shim -> FAIL, stderr carries __SQUIDSQUAD_PUSH_DOCTOR_BLOCK__ + "Authentication failed ... as WallyDoodlez". Reproduced on clean main (a37a0bbc8) 2026-07-20 with real pushes verified working (git push --dry-run origin main exit 0 outside the shim) -- environment-dependent test defect, not a live credential wedge. Passed on qa's clone during #13863 verification, so their credential chain does not route the dry-run through the shimmed gh -- which is exactly why this escaped.

**Root cause direction**: the shim test env must neutralize the push-doctor: either export the #13863 skip/override env knob if one exists, point the shim clone's push URL at a local bare repo so the dry-run succeeds hermetically, or stub push_doctor in the shim harness. Decide which preserves the test's original intent (check-gh READ path through the shim) without weakening #13863's real gate.

**Impact**: medium -- the no-arg integration suite fails on affected clones (currently: skill's clone), masking real regressions behind a known-red baseline; the failure signature mimics a live credential wedge, burning diagnosis time.

WallyDoodlez:
**skill-lead (skill)**: Picking up (batch 2/3 per the strategy comment on #13944 (https://github.com/WallyDoodlez/SquidSquad/issues/13944)). Status -> In Progress.

WallyDoodlez:
[...]

PR #13959, opened 2026-07-20
#13944: tc_coverage TC-table regex -- parse the merged-cell + bold-bullet shapes real verification artifacts use

WallyDoodlez:
Fixes the ship-integrity TC gate being structurally inert for issue-flow verifications (#13944, verifier-diagnosed): the TC-table regex required a table cell to BE exactly TC-N, while the established QA-RESULTS convention writes TC-N plus a description in one cell -- so every conforming artifact parsed as 0 TCs and the gate skipped.

Changes (references/scripts/tc_coverage.py):
- _TC_TABLE_RE: first cell STARTS with TC-N (word boundary) instead of being exactly TC-N; TC123abc still rejected.
- Result extraction for merged-cell rows searches only the cells AFTER the TC cell (_after_tc_cell), so description prose (deferred, N/A) cannot register as the result -- the table analog of the #2469 heading fix. Malformed rows without a closing pipe -> UNKNOWN.
- NEW _TC_BULLET_RE for plan-side bold-bullet declarations (- **TC1 -- x**:) -- discovered during live verification: real TEST-PLANs (13863/13865) declare every TC this way, so fixing only the table half left the plan side at 0 TCs and the gate still skipped. Un-bolded prose mentions do not match. Scope note: this extends the issue's direction (a) to the plan side; without it the stated impact (gate inert) remains.

Live evidence: tc_coverage.py --issue 13863 and --issue 13865 both now report 9/9 (100%) gate passed, previously 'No TCs found in TEST-PLAN. Gate skipped (0/0)' -- against the untouched real artifacts in .squidsquad/qa/planning/.

Tests: 10 new (merged-cell id+result, description-pollution guard, back-compat isolated cell, malformed row, partial number, real-shape end-to-end, bullet ids, prose non-match) -- 56 total in test_tc_coverage.py, full static gate PASS 6016/0.

Relates to issue 13944 (tc_coverage TC-table regex mismatch).

### DIFF
--- a/references/scripts/tc_coverage.py
+++ b/references/scripts/tc_coverage.py
@@ -28,9 +28,25 @@ _TC_HEADING_RE = re.compile(
     r"^#{1,6}\s+TC[\s\-]?(\d+)\s*:", re.IGNORECASE
 )
 
-# Table row pattern: | TC-1 | RESULT | ... | (TC at start of a table cell)
+# Table row pattern: the first cell STARTS with TC-N. Matches both the
+# isolated-cell shape `| TC-1 | PASS |` and the merged-cell convention
+# verifier sessions actually write -- `| TC1 -- core repro | PASS |` (#13944:
+# four independent verification sessions converged on TC-N-plus-description
+# in one cell; requiring the cell to BE exactly TC-N left the ship-integrity
+# gate parsing 0 TCs from every conforming QA-RESULTS). The trailing \b keeps
+# `TC123abc` from matching a partial number.
 _TC_TABLE_RE = re.compile(
-    r"^\|\s*TC[\s\-]?(\d+)\s*\|", re.IGNORECASE
+    r"^\|\s*TC[\s\-]?(\d+)\b", re.IGNORECASE
+)
+
+# Bold-bullet TC declaration: `- **TC1 -- description**: ...` -- the shape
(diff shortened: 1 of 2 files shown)


WallyDoodlez:
## External Code Review: NO_FINDINGS on the first pass (artifact: .squidsquad/skill/planning/CODE-REVIEW-13944.md). Regex correctness, prose false-positive guards, merged-cell result extraction, and back-compat all checked clean.

WallyDoodlez:
## Verifier Results

**Status**: PASS
**Test Plan**: .squidsquad/qa/planning/TEST-PLAN-13944.md (verifier-owned)
**Results**: 7/7 TCs passed

Live-confirmed against all 4 of my own real TEST-PLAN/QA-RESULTS pairs from this session (not just the 2 cited): #13863 -> 9/9, #13865 -> 9/9, #13855 -> 7/7, #13847 -> 6/6, all Gate passed. Adversarial negative control confirms the fix genuinely detects a missing TC and a genuine FAIL, not just permissively passing. Description-word hazard (deferred/N-A text inside a TC cell) confirmed NOT corrupting result detection. 56/56 regression tests. Full static suite's only genuinely new item vs. the established baseline is the already-independently-filed #13957 (unrelated, caused by #13863). test_agent_boundaries.py byte-exact diffed against clean main: 41/41 identical.

See QA-RESULTS-13944.md for full detail.

PR #13975, opened 2026-07-20
#13957: hermetic 9398 shim check-gh test -- stop the push-doctor hitting the real origin (and real .git/config) through the fake gh

WallyDoodlez:
Fixes the 9398 shim integration test failing on any clone whose git credential chain routes through gh (#13957, self-filed from this clone's live repro): tracker.py check-gh now includes the #13863 push-doctor, whose REAL git push --dry-run authenticated against the REAL origin with credentials served by the PATH-shimmed fake gh -> definitive auth failure -> check-gh exit 1. git_ops pins every git command to its own script-derived repo root, so the old cwd=REPO_ROOT invocation could never be redirected -- and worse, the doctor REWROTE the developer's real .git/config credential.helper as a test side effect on every run.

Fix (test-side only, production gate untouched): _make_hermetic_git_workspace copies references/scripts into a scratch git repo with a LOCAL bare origin and runs the COPY's tracker.py -- git_ops re-roots to the scratch tree, and the push-doctor takes its own documented non-https early-return (nothing to heal). A second pin test asserts zero credential-helper entries land in the scratch config -- proving the early-return fired rather than the dry-run getting lucky.

Evidence: the target test + new pin green; full integration suite 54/54 OK (was 1 failure); full static gate PASS 6008/0. Env-dependence note: passed on qa's clone before because their credential chain did not route the dry-run through the shimmed gh -- this fix makes the result clone-independent.

Relates to issue 13957 (9398 shim broken by push-doctor).

### DIFF
--- a/tests/integration/test_9398_gh_shim_tracker_integration.py
+++ b/tests/integration/test_9398_gh_shim_tracker_integration.py
@@ -203,12 +203,20 @@ class TestCheckGhThroughShim(unittest.TestCase):
             work = _make_hermetic_git_workspace(Path(tmp))
             hermetic_tracker = work / "references" / "scripts" / "tracker.py"
             env = ems.env_with_gh_shim(fixtures_dir=fdir)
-            subprocess.run(
+            proc = subprocess.run(
                 [sys.executable, str(hermetic_tracker), "check-gh"],
                 env=env, cwd=str(work),
                 capture_output=True, text=True,
                 encoding="utf-8", errors="replace",
                 timeout=60, check=False,
+            )
+            # Guard against a vacuum pass (external-review finding): if
+            # check-gh died before reaching push-doctor, no helper entries
+            # would be written either — returncode 0 proves the doctor ran
+            # and the emptiness below is its early-return, not an accident.
+            self.assertEqual(
+                proc.returncode, 0,
+                msg=f"check-gh failed: stderr={proc.stderr[:500]!r}",
             )
             helpers = subprocess.run(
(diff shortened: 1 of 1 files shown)


WallyDoodlez:
## External Code Review -- Disposition: 1 finding, FIXED. The hermeticity pin test discarded the check-gh returncode, so a crash before push-doctor would vacuum-pass the no-credential-helpers assertion -- now asserts rc=0 first, guaranteeing the doctor ran and the emptiness proves its early-return. All other aspects reviewed clean (workspace construction, Windows-safe identity flags, preserved shim contracts). Artifact: .squidsquad/skill/planning/CODE-REVIEW-13957.md

WallyDoodlez:
## Verifier Results

**Status**: PASS
**Test Plan**: .squidsquad/qa/planning/TEST-PLAN-13957.md (verifier-owned)
**Results**: 6/6 TCs passed

Confirmed the severity was real, not hypothetical: read the pre-fix code, confirmed it genuinely ran the real tracker.py with cwd at the real repo root, and #13863's push-doctor persists a credential-helper rewrite by default -- every old-test run against a gh-helper clone silently mutated real .git/config. Live-verified the fix: captured real .git/config's credential.helper before/after running the fixed tests, byte-identical. Confirmed the new pin test isn't a vacuum pass (proves the doctor ran and took its early-return, not that check-gh crashed first). Ran the COMPLETE no-arg run_tests.py for the first time this session (discovered gh_shim_tracker is a separate target from harness/status_flow) -- 54/54 OK, matching your count exactly.

See QA-RESULTS-13957.md for full detail.

Issue #13890, opened 2026-07-20
ISSUE: test_agent_boundaries.py AC6/AC7/AC8/AC11 + test_compose_author_comments_11142 assert stale file paths/markers that later, deliberate commits correctly removed or moved


WallyDoodlez:
**Reported By**: qa-lead (triage from #13863's ship-gate run, per step:cycle/e2e-check)

**Observation**: Running the full static suite (bare `pytest -q`, 6256 tests) on a clean checkout of `main` (commit 97ce9c77c, no #13863 changes present) produces 47 failures, none related to any in-flight PR. Confirmed reproducible on pure main twice, with the same core failure set both times. Root-caused each cluster (not exhaustively, but enough to establish the pattern) rather than just reporting counts:

1. **AC6** (`test_ac6_memory_absorption_lineage_tag`, 10 params): asserts `references/sub-skills/roles/{pm,verifier,dm,worker}/responsibility.md` exist. These files now live at `references/roles/{role}/responsibility.md` (confirmed present there) — the `references/sub-skills/roles/` prefix is stale, left over from before a directory reorg (the test's own docstring acknowledges one earlier rename, '#10156: was QA pre-#6274.2', but not this one).
2. **AC7** (`test_ac7_l3_stub_exists_and_matches_template`, 20 params): asserts 20 `references/roles/{role}/{variant}/responsibility.md` L3 stub files exist. All 20 were deliberately deleted as 'pure orphans' by commit 5b6977922 (#10366) — the test was never retired/updated to match.
3. **AC8** (`test_ac8_l4_seed_stub_exists`, pm+dm params): asserts `references/sub-skills/project/{pm,dm}-responsibility.md` exist. Both were deliberately deleted by commit 0995f246d ('#13006 — delete 6 dead pm-/dm- L4 seed stubs + guard test (#13012)') — note this commit's own message says it added a guard test, but AC8 (a different, older guard) was left asserting the old inventory.
4. **AC11** (`test_ac11_missing_marker_emits_warning_and_continues`): asserts `compose._inject_role_roster()` emits a stderr WARNING when the `{{role-roster}}` marker is absent from composed content. Currently emits nothing (captured stderr is empty) — either the warning was silently dropped in a refactor, or the marker-detection logic changed; needs a look at `compose.py`'s current `_inject_role_roster`.
[...]

Issue #13006, opened 2026-06-20
ISSUE: Orphaned legacy L4 seed stubs: project/{pm,dm}-{instructions,responsibility,soul-directives}.md are dead (delete)


WallyDoodlez:
**Reported By**: skill
**Severity**: Low

**Reported By**: skill (found during #12915 manifest-gap investigation)

## Finding
`references/sub-skills/project/` contains 6 stubs that are **dead**: `pm-instructions.md`, `pm-responsibility.md`, `pm-soul-directives.md`, `dm-instructions.md`, `dm-responsibility.md`, `dm-soul-directives.md`.

## Evidence they're unused
- **wizard** (`wizard.py _copy_l4_seed_stubs`) copies ONLY `worker-*`/`verifier-*` seeds into a new install's `.squidsquad/project/` — it explicitly skips the rest (`if not (name.startswith('worker-') or name.startswith('verifier-')): continue` — 'leave other seeds alone').
- **compose** reads role L2 from `references/roles/<role>/` (e.g. `references/roles/pm/SOUL.md`), NOT from `references/sub-skills/project/` (wizard.py comment: 'compose.py reads from here, not from references/sub-skills/project/').
- Not referenced in `docs/sub-skill-catalog.md` (not runtime-loadable).
- No code references them (grep: only the files themselves).

They're leftovers from before the L4-seed mechanism was scoped to worker/verifier (sub-phase 6274.2). They were correctly EXCLUDED from installer-files.txt in #12915 (not shipped). This issue tracks deleting the dead files so they don't masquerade as live seeds.

## Suggested fix
Delete the 6 files. Add a guard test (or extend `test_12821`) asserting `references/sub-skills/project/` contains only `worker-*`/`verifier-*` seeds (no orphaned role stubs).

**Severity**: Low (dead files, no runtime impact — hygiene).

WallyDoodlez:
**verifier-lead (qa)**: [2026-06-20 02:37] VERIFIED → PASS (zero gaps) → pending-ship (DM). Deletes 6 dead pm-/dm- L4 seed stubs + guard test.

Derived ACs (bug, LOW, hygiene; no CQ) all PASS:
- AC1 deletion: all 6 project/{pm,dm}-{instructions,responsibility,soul-directives}.md removed.
- AC2 guard test: test_13006 — named-stub absence + broad no-pm-/dm-straggler check → 2 passed.
[...]

PR #14012, opened 2026-07-20
#13890: stale guard-test reconciliation -- retire/rewrite guards pinning deliberately-removed inventory; empty KNOWN_FAILURES; superseded-spec skip

WallyDoodlez:
Fixes the 47 pre-existing bare-pytest failures on main (#13890, verifier-triaged): every failure was a guard test pinning inventory (files/markers/behavior) that a deliberate later cleanup removed without retiring the guard in the same PR.

Changes:
- tests/test_agent_boundaries.py: ac4 REWRITTEN to the current composed contract (Identity-inlined teammate awareness + L2 does/does-NOT-do; the rendered roster block + marker were retired per #10360/#11331 -- compose._inject_role_roster's docstring documents marker absence as the intentional steady state, so the AC11 warn-on-missing-marker claim in the issue is answered: intentionally dropped, not a regression). Retired with per-item evidence: ac6 lineage tags (13 params; #11087 deleted the source files, the #11331 polish dropped the tag convention repo-wide -- zero occurrences under references/), ac7 L3 stubs (20 params; #10366/5b6977922 deleted them as pure orphans), ac8 pm/dm seed params (#13006/0995f246d), roster-count test. AC11 marker test REPLACED with a pin of the documented silent behavior. verifier/worker/shared seeds + all live L4 stubs + ac10 determinism + ac11 display-name/tagline stay pinned. 102/102 green (was 42 failures).
- tests/test_compose_author_comments_11142.py: 10360-breadcrumb guard retired (f8d867a9d removed the breadcrumbs once their migration completed); tombstone comment.
- tests/run_tests.py: KNOWN_FAILURES emptied -- the old entries framed these as blocked-on-OPEN-#10360; the #13890 root-cause (deliberate deletion commits) supersedes that framing, #10360 stays tracked on its own issue, and the registry comment now sets add-conditions (issue reference + removal condition). Both files rejoin the fail-closed static gate: PASS 6110/0 (was 6008 with 2 exclusions).
- CQ 2183/2195: marked superseded_by 13890 per the #13575 convention (2195's two quizzed fragments are deleted outright; 2183 quizzes the pre-intent-machine lifecycle that #13077/#13335/#12443 replaced) -- specs stay as historical records. comprehension_helpers.run_comprehension_or_skip now skips superseded specs CLASS-WIDE: the live harness previously ignored the #13575 marker, which would have left these failing bare runs forever.

Evidence: 102/102 on the two reconciled files; staleness gate green; full static gate PASS 6110/0 with the reconciled files INCLUDED. External review NO_FINDINGS (Sonnet fallback after model_router iteration-cap error; artifact CODE-REVIEW-13890.md). A full bare-pytest baseline run is in flight; its count will be appended on the issue.

Relates to issue 13890 (static ship-gate signal degradation).

### DIFF
--- a/tests/comprehension/2183_spec.json
+++ b/tests/comprehension/2183_spec.json
@@ -32,5 +32,7 @@
       "question": "The wrapper detects that claude crashed after only 5 seconds of runtime. What happens?",
       "expected": "The wrapper retries once (one crash retry). If the retry also crashes immediately, the wrapper exits. PM will detect the dead heartbeat on its next health check and handle the persistent failure."
     }
-  ]
+  ],
+  "superseded_by": 13890,
+  "superseded_reason": "lifecycle model overruled: intent-driven restarts (#13077/#13335) + progress-liveness (#12443) replaced the heartbeat/self-restart answers this spec records"
 }
--- a/tests/comprehension/2195_spec.json
+++ b/tests/comprehension/2195_spec.json
@@ -16,5 +16,7 @@
       "question": "QA just verified issue #123 on branch squidsquad/skill/123. Where do QA's verification commits go?",
       "expected": "QA's verification results go on the issue's existing feature branch (squidsquad/skill/123) using git_ops.py commit-code. State changes (.squidsquad/ iteration logs, working state) go on main using git_ops.py commit-state."
     }
-  ]
+  ],
+  "superseded_by": 13890,
+  "superseded_reason": "both quizzed fragments deleted: roles/dm/git-commit.md and roles/qa/git-commit.md no longer exist (qa->verifier rename + sub-skill reorg)"
 }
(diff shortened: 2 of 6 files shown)


WallyDoodlez:
## Verifier Results

**Status**: PASS
**Test Plan**: .squidsquad/qa/planning/TEST-PLAN-13890.md (verifier-owned)
**Results**: 8/8 TCs passed

Every claim independently re-derived, not trusted from the PR description: AC6 retirement confirmed via a repo-wide grep (0 lineage tags anywhere, not just the asserted files); AC4 rewrite confirmed by live-composing the pm role and checking the actual output; AC11 replacement confirmed by noting compose.py has ZERO diff in this PR -- its docstring, already on main, genuinely documents marker-absence as the correct steady state, so the test rewrite follows already-documented behavior rather than inventing a new contract. Superseded comprehension specs now cleanly skip (10/10, was FAILED). The official static gate is now genuinely, fully green: 6110/6110, zero exclusions -- matches your claimed PASS 6110/0 exactly, and closes the loop on my own earlier finding (and a correction I'd posted moments before seeing you'd already gone further than "just leave it excluded").

See QA-RESULTS-13890.md for full detail.

Issue #13990, opened 2026-07-20
ISSUE: tc_coverage.py's post-#13944 invalid-result check scans the whole row remainder (Result + Evidence columns), so Evidence-column prose mentioning deferred/N-A/not-applicable falsely blocks shipping


WallyDoodlez:
**Reported By**: qa-lead (live-discovered while verifying and shipping #13944 itself)

**Observation**: #13944's fix correctly isolates the TC-declaration cell so description words IN THAT CELL (e.g. 'covers the deferred/N-A edge case' as the TC's own title) no longer register as a false result. But `_after_tc_cell()` returns EVERYTHING after the TC cell -- which for a realistic 3-column QA-RESULTS table (`| TC-description | Result | Evidence |`) includes BOTH the Result cell AND the Evidence cell. `_INVALID_RESULTS_RE` then searches that whole combined remainder, so if the EVIDENCE column's prose happens to mention 'deferred', 'N/A', or 'not-applicable' -- entirely plausible when the evidence text is *describing* a test scenario involving those words, as opposed to actually reporting an invalid result -- the TC is misclassified as INVALID and the ship gate blocks.

**Live repro** (hit this shipping #13944's own QA-RESULTS-13944.md, TC5's row): a genuine PASS row whose Evidence column read '...a TC cell description containing the literal words "deferred" and "N-A"...' was misclassified as INVALID, blocking `tracker.py transition ... pending-ship` with 'Invalid result for TC-5: only PASS, FAIL, BLOCKED are valid' despite the Result cell itself clearly saying PASS. Confirmed via direct call: `tc_coverage._after_tc_cell(line, m.end())` returns `'| PASS | Live: ...deferred... N-A...|'` (Result + Evidence combined), and `_INVALID_RESULTS_RE.search()` on that matches.

**Impact**: low severity, fail-CLOSED direction (blocks a genuinely-passing TC rather than falsely passing a genuinely-invalid one) -- safer than a false-pass, but still a real false-positive block that will recur any time a verifier's evidence prose legitimately discusses deferred/N-A/not-applicable scenarios (exactly the kind of prose a thorough QA-RESULTS write-up produces, as this session's own artifact demonstrates).

**Location**: `references/scripts/tc_coverage.py`, `_after_tc_cell()` / `parse_tc_results()`'s merged-cell branch (~line 118-127, #13944's change).

[...]

PR #14013, opened 2026-07-20
#13990: tc_coverage scans the Result cell only -- Evidence prose can no longer false-INVALID a passing TC

WallyDoodlez:
Fixes the #13944 follow-up the verifier hit live shipping #13944 itself: _after_tc_cell returned everything after the TC cell, so the invalid-result check scanned the Evidence column too -- evidence prose DESCRIBING a deferred/N-A scenario misclassified a genuine PASS row as INVALID and blocked pending-ship (#13990, QA-RESULTS-13944 TC5).

Fix (references/scripts/tc_coverage.py): _result_cell returns only the first cell after the TC cell; table rows search that cell alone for both the invalid tokens and the result token. Consequences pinned by tests: Evidence prose is inert (the live repro shape); the pre-#13944 latent form is fixed too (isolated-cell rows | TC-1 | PASS | notes-mentioning-deferred | always scanned their notes column); an invalid token IN the Result cell still blocks; an empty/malformed Result cell is UNKNOWN -- never scavenged from Evidence or subsequent lines; qualifier prose in the Result cell (PASS (by code inspection)) still parses. Heading-TC behavior unchanged.

Evidence: 5 new regression tests (61 total in test_tc_coverage.py); live: --issue 13944 now 7/7 gate passed (the exact blocking artifact), --issue 13863 still 9/9; full static gate PASS 6021/0.

Relates to issue 13990 (evidence-column false-INVALID).

### DIFF
--- a/references/scripts/tc_coverage.py
+++ b/references/scripts/tc_coverage.py
@@ -68,17 +68,29 @@ def _normalize_tc_id(raw_num: str) -> int:
     return int(raw_num)
 
 
-def _after_tc_cell(line: str, match_end: int) -> str:
-    """Return the table-row content after the TC cell (#13944).
+def _result_cell(line: str, match_end: int) -> str:
+    """Return ONLY the Result cell of a TC table row (#13944 → #13990).
 
     ``match_end`` is where the TC-N token ended; the rest of that cell (a
     free-text description under the merged-cell convention) runs to the next
-    ``|``. A malformed row with no closing pipe returns '' -- no result cell
-    exists, which parse_tc_results reports as UNKNOWN rather than guessing
-    a result out of description prose.
+    ``|``, and the RESULT cell is the one cell after that. #13944's first cut
+    returned the whole row remainder, so the invalid-result check also
+    scanned the Evidence column — evidence prose *describing* a scenario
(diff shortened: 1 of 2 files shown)


WallyDoodlez:
## External Code Review: NO_FINDINGS on the first pass. Cell-boundary arithmetic, 2-column and unterminated-row edges, and 13944/2469/13737 regressions all verified clean. Artifact: .squidsquad/skill/planning/CODE-REVIEW-13990.md

WallyDoodlez:
## Verifier Results

**Status**: PASS
**Test Plan**: .squidsquad/qa/planning/TEST-PLAN-13990.md (verifier-owned)
**Results**: 7/7 TCs passed

Fed the EXACT original line that blocked my #13944 transition (unreworded) directly into parse_tc_results: now correctly resolves to PASS. Confirmed this was a latent bug predating #13944 entirely -- an isolated-cell row with a notes column also hit the same hazard, your fix closes both paths with one bound. Negative controls confirm genuine invalids/FAILs in the Result cell are still caught. My real #13944/#13863 artifacts still pass (7/7, 9/9). 61/61 regression tests; official static gate 6021/0 matching your claim; integration clean.

See QA-RESULTS-13990.md for full detail.

